PR #13, opened 2025-06-23
Add action-perception diagram

tristan-deep:
Added a diagram to about page. Thanks to @OisinNolan for making it.


docs/source/about.rst
```diff
@@ -5,6 +5,37 @@ About
...
+     />
+     <!-- Light mode image -->
+     <img
+       src="../_static/zea_perception_action-Light.svg"
+       alt="Closing the action-perception loop in ultrasound imaging"
+       style="display: none; width: 80%; padding-bottom: 1em;"
+       class="only-light"
+     />
+     <div style="text-align: center; font-style: italic; color: var(--color-foreground-secondary, #666);">
+       Conceptual overview of the action-perception loop in ultrasound imaging.
```
OisinNolan, line 26: Nit: maybe we should say:
'High-level overview of 'an' ultrasound perception-action loop implemented in `zea`'
Or something along those lines -- I think using 'the' makes it sound like this is the only thing that is meant by perception-action loop, when actually it is just one instance of a perception-action loop

OisinNolan (approved):
Nice, thanks for adding this!

Issue #14, opened 2025-06-24
Diffusion module needs tensorflow


vincentvdschaft:
When I import zea in a new environment after installing only zea and JAX I get an error:
```bash
site-packages/zea/models/diffusion.py", line 7, in <module>
    import tensorflow as tf
ModuleNotFoundError: No module named 'tensorflow'
```
This is because of this import
https://github.com/tue-bmd/zea/blob/0dd97225001e39f017c35b7fae392e1cd3f28581/zea/models/diffusion.py#L7

The import is only used in the line
https://github.com/tue-bmd/zea/blob/0dd97225001e39f017c35b7fae392e1cd3f28581/zea/models/diffusion.py#L309

Can this dependency be removed or done through Keras?

tristan-deep:
Ah, I thought I fixed that in https://github.com/tue-bmd/zea/commit/9d9b897cb6420a5dc373a4904468b85e6e79995d, but apparently not. We can fix it with a similar lazy import, such that it will only error out when people try to train the diffusion model (i.e. use the `train_step` method).

PR #16, opened 2025-06-24
Fix Diffusion module needs tensorflow #14

tristan-deep:
Fixes imports.


tristan-deep:
Let me quickly add a test for this as well. I think these kind of imports sneak in easily.

PR #18, opened 2025-06-24
Reuse previous docker builds to speed-up CI

benluijten:
Docker images should only be rebuild if pyproject.toml, poetry.lock or Dockerfile change. 

This PR adds a check for this, and retags the image from the previous commit if no changes are detected. the first run in a PR will always trigger a build. Also, if no image was found on zeahub, a build is triggered.


tristan-deep:
Should also change when [`motd.sh`](https://github.com/tue-bmd/zea/blob/main/motd.sh) changes I believe.

tristan-deep (approved):
Good change, will save us some time 🚀. lgtm!

PR #21, opened 2025-06-25
Tensor caching for parameters

benluijten:
Added a cache that stores tensor versions of parameters/properties in the Parameter class. It is now also possible to exclude parameters from to_tensor() conversion, or only include specific keys (replaces compute_keys).

### DIFF
--- a/zea/internal/parameters.py
+++ b/zea/internal/parameters.py
@@ -396,12 +396,12 @@ class Parameters(ZeaObject):
         property_keys = set(self._properties_with_dependencies)
         all_keys = param_keys | property_keys | set(self._properties)
 
-        if include is not None and include != "all":
+        if include == "all":
+            keys = all_keys
+        elif include is not None:
             keys = set(include).intersection(all_keys)
-        else:
-            keys = set(all_keys)
-        if exclude is not None:
-            keys = keys - set(exclude)
+        elif exclude is not None:
+            keys = all_keys - set(exclude)
 
         tensor_dict = {}
         # Convert parameters and computed properties to tensors
@@ -410,7 +410,7 @@ class Parameters(ZeaObject):
             try:
                 val = getattr(self, key)
             except MissingDependencyError as exc:
-                if include == "all":
+                if include == "all" or exclude is not None:
                     # If we are including all, we can skip this key
                     continue
                 else:
--- a/tests/test_parameters.py
+++ b/tests/test_parameters.py
@@ -195,6 +195,20 @@ def test_to_tensor_includes_all(dummy_params: Parameters):
     )
 
 
+def test_to_tensor_excludes(dummy_params: Parameters):
+    """Test that to_tensor excludes specified keys."""
+    # Exclude computed1 and param2
+    tensors = dummy_params.to_tensor(exclude=["computed1", "param2"])
+    assert "computed1" not in tensors
+    assert "param2" not in tensors
+    # Should still include other params and computed properties
+    for key in ["param1", "param3", "param4", "computed2", "computed3"]:
+        assert key in tensors
+
+    # Exclude a non-existent key, should not raise error
+    dummy_params.to_tensor(exclude=["non_existent"])
+
+
 def test_to_tensor_partial_computed_subset(dummy_params):
     """Test that to_tensor only computes the requested subset."""
     # Access no computed properties yet


benluijten:
Small benchmark:

```python
import time

import numpy as np

# disable zea logging if too verbose
from zea import log
from zea.probes import Verasonics_l11_5v
from zea.scan import Scan

# log.logger.setLevel("WARNING")

def make_scan():
    """
    Create a minimal but non-trivial Scan instance with default parameters.
    """
    probe = Verasonics_l11_5v()
    return Scan(
        Nx=128,
        Nz=256,
        xlims=(-0.02, 0.02),
        zlims=(0.0, 0.06),
        center_frequency=6.25e6,
        sound_speed=1540.0,
        sampling_frequency=25e6,
        n_el=128,
        n_tx=11,
        polar_angles=np.linspace(-0.1, 0.1, 11),
        azimuth_angles=np.linspace(-0.2, 0.2, 11),
        t0_delays=np.zeros((11, 128)),
        tx_apodizations=np.ones((11, 128)),
        focus_distances=np.ones(11),
        initial_times=np.zeros(11),
        time_to_next_transmit=np.ones((1, 11)),
    )


def time_to_tensor(scan, n_runs=10, **to_tensor_kwargs):
    """
    Measure the average and standard deviation of scan.to_tensor() over n_runs.
    Accepts any keyword args supported by to_tensor(), e.g., exclude.
    """
    times = []
    for _ in range(n_runs):
        t0 = time.perf_counter()
        scan.to_tensor(**to_tensor_kwargs)
        t1 = time.perf_counter()
        times.append(t1 - t0)
    return np.mean(times)


def main():
    scan = make_scan()

    # Case 1: initial warm-up (uncached)
    t0 = time.perf_counter()
    scan.to_tensor()
    t1 = time.perf_counter()
    print("[Case 1] Warm-up (uncached, all outputs):")
    print(f"{(t1 - t0)*1000:.3f} ms")

    # Reset cache for fresh timing
    scan._tensor_cache = {}

    # Case 2: first run populates cache
    mean1 = time_to_tensor(scan, n_runs=1)
    print("[Case 2] Populate cache (first run):")
    print(f"{mean1*1000:.3f} ms")

    # Case 3: cached runs
    mean2 = time_to_tensor(scan, n_runs=10)
    print("[Case 3] Cached runs:")
    print(f"{mean2*1000:.3f} ms")

    # Case 4 & 5: change single parameter Nx and time
    scan.Nx = 64
    mean3 = time_to_tensor(scan, n_runs=1)
    print("[Case 4] After updating Nx=64 (invalidate cache):")
    print(f"{mean3*1000:.3f} ms")
    mean4 = time_to_tensor(scan, n_runs=10)
    print("[Case 5] Using cached to_tensor after Nx change:")
    print(f"{mean4*1000:.3f} ms")

    # Case 6 & 7: change single parameter sound_speed and time
    scan.sound_speed = 1500.0
    mean5 = time_to_tensor(scan, n_runs=1)
    print("[Case 6] After updating sound_speed=1500.0 (invalidate cache):")
    print(f"{mean5*1000:.3f} ms")
    mean6 = time_to_tensor(scan, n_runs=10)
    print("[Case 7] Using cached to_tensor after sound_speed change:")
    print(f"{mean6*1000:.3f} ms")

    # Case 8 & 9: update multiple parameters at once and time
    multi_params = {"Nx": 128, "Nz": 256, "xlims": (-0.03, 0.03), "zlims": (0.0, 0.07)}
    scan.update(**multi_params)
    mean7 = time_to_tensor(scan, n_runs=1)
    print(f"[Case 8] After updating {multi_params} (invalidate cache):")
    print(f"{mean7*1000:.3f} ms")
    mean8 = time_to_tensor(scan, n_runs=10)
    print("[Case 9] Using cached to_tensor after multiple updates:")
    print(f"{mean8*1000:.3f} ms")

    # Case 10 & 11: change sound_speed randomly and exclude 'pfield'
    new_ss = np.random.uniform(1400, 1600)
    scan.sound_speed = new_ss
    mean9 = time_to_tensor(scan, n_runs=10, exclude=["pfield"] )
    print(f"[Case 10] to_tensor exclude=['pfield'] after sound_speed={new_ss:.2f}:")
    print(f"{mean9*1000:.3f} ms")
    mean10 = time_to_tensor(scan, n_runs=10, exclude=["pfield"] )
    print("[Case 11] Cached to_tensor exclude=['pfield']:")
    print(f"{mean10*1000:.3f} ms")

if __name__ == "__main__":
    main()
```

output:
```
[Case 1] Warm-up (uncached, all outputs):
875.602 ms
[Case 2] Populate cache (first run):
8.433 ms
[Case 3] Cached runs:
0.007 ms
[Case 4] After updating Nx=64 (invalidate cache):
0.010 ms
[Case 5] Using cached to_tensor after Nx change:
0.011 ms
[Case 6] After updating sound_speed=1500.0 (invalidate cache):
0.526 ms
[Case 7] Using cached to_tensor after sound_speed change:
0.006 ms
[Case 8] After updating {'Nx': 128, 'Nz': 256, 'xlims': (-0.03, 0.03), 'zlims': (0.0, 0.07)} (invalidate cache):
0.945 ms
[Case 9] Using cached to_tensor after multiple updates:
0.007 ms
[Case 10] to_tensor exclude=['pfield'] after sound_speed=1494.49:
0.051 ms
[Case 11] Cached to_tensor exclude=['pfield']:
0.006 ms
```

tests/test_ops_infra.py
```diff
@@ -398,7 +398,9 @@ def test_pipeline_with_scan_probe_config():
     result = pipeline(**parameters, x=2, y=3)
 
     assert "z" in result
-    assert "n_tx" in result  # Check if we parsed the scan object correctly
+    # FIX: The following assertion is commented out because we only convert self.valid_keys
+    # from Scan to tensors. In this case, this is only x and y, not n_tx.
+    # assert "n_tx" in result  # Check if we parsed the scan object correctly
```
benluijten, line 403: Any suggestions on how to address this? Previously we apparently did convert n_tx from the Scan class, even though it was not part of self.valid_keys(), only 'x' and 'y' are in this case. Remove the assertion?

zea/ops.py
```diff
@@ -877,10 +883,7 @@ def prepare_parameters(
             assert isinstance(scan, Scan), (
                 f"Expected an instance of `zea.scan.Scan`, got {type(scan)}"
             )
-            scan_dict = scan.to_tensor(
-                compute_missing=True,
-                compute_keys=self.valid_keys,
-            )
+            scan_dict = scan.to_tensor(include=self.valid_keys)
```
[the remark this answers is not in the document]
tristan-deep: Good point, I'm not sure. I think we can change that. Also we can make the config line similar to probe and scan, so the exact same thing happens to all.

zea/internal/parameters.py
```diff
@@ -322,37 +328,63 @@ def _resolve_dependency_tree(self, name, failed=None):
...
+        Get all properties of this class
         """
-        Convert all parameters (and optionally computed properties) to tensors.
+        properties = set()
+        for name, attr in self.__class__.__dict__.items():
+            if isinstance(attr, property) and hasattr(attr.fget, "_dependencies"):
+                properties.add(name)
+        return properties
+
+    def to_tensor(self, include="all", exclude=None, compute=True, skip_missing=True):
```
tristan-deep, line 341: I find these arguments a bit confusing. Could you extend the docstring a bit, or give some examples there? Do we really need all four arguments? I can appreciate the flexibility they add, but I wonder in what scenarios they would be useful (if so, they would be good examples, or used somewhere in the codebase).

zea/internal/parameters.py
```diff
@@ -322,37 +328,63 @@ def _resolve_dependency_tree(self, name, failed=None):
...
+        tensor_dict = {}
+        # Convert parameters and computed properties to tensors
+        for key in keys:
+            # Get the value from params or computed properties
+            try:
+                val = getattr(self, key)
+            except AttributeError as e:
+                if skip_missing:
                     continue
-                attr = getattr(self.__class__, name)
```
tristan-deep, line 341: I'm not sure what skip missing exactly does (see previous comment), but would be good to add a test for this, as it seems a crucial part of the pipeline / parameter handling.

tristan-deep (changes requested):
Nice update! Should bring some good speedups with parameter handling 🚀

Unfortunately we don't have tests for the example notebooks yet, but the [pipeline notebook](https://zea.readthedocs.io/en/latest/notebooks/pipeline/zea_pipeline_example.html) is now failing. I have now added a test on this branch that catches that error, maybe you can take a look at it. Basically, it didn't catch that changing selected_transmits with scan.set_transmits would invalidate the cached tensors. 

I think, since you now added quite some logic around to_tensor, it would be nice to add a separate test for it, so we know what behavior is expected, and it will not break in the future. Seems there is not a lot of testing for to_tensor. Especially now the complexity of to_tensor has grown with the multiple arguments and flexibility it supports. 

As you hint to yourself, we probably also want to have a look how to make to_tensor more consistent across probe, scan and config. But maybe in a later PR :).

benluijten:
> Nice update! Should bring some good speedups with parameter handling 🚀
> 
> **Unfortunately we don't have tests for the example notebooks yet, but the [pipeline notebook](https://zea.readthedocs.io/en/latest/notebooks/pipeline/zea_pipeline_example.html) is now failing. I have now added a test on this branch that catches that error, maybe you can take a look at it. Basically, it didn't catch that changing selected_transmits with scan.set_transmits would invalidate the cached tensors.**
> 
> I think, since you now added quite some logic around to_tensor, it would be nice to add a separate test for it, so we know what behavior is expected, and it will not break in the future. Seems there is not a lot of testing for to_tensor. Especially now the complexity of to_tensor has grown with the multiple arguments and flexibility it supports.
> 
> As you hint to yourself, we probably also want to have a look how to make to_tensor more consistent across probe, scan and config. But maybe in a later PR :).
> 
> **BTW; this bug might be solved in #23. if the dependency resolution was broken and you rely on it with to_tensor, this could fix it already.**

to verify, this bug was not introduced in this PR right? I noticed it on the main branch as well, see #28.

tristan-deep:
> > Nice update! Should bring some good speedups with parameter handling 🚀
> > **Unfortunately we don't have tests for the example notebooks yet, but the [pipeline notebook](https://zea.readthedocs.io/en/latest/notebooks/pipeline/zea_pipeline_example.html) is now failing. I have now added a test on this branch that catches that error, maybe you can take a look at it. Basically, it didn't catch that changing selected_transmits with scan.set_transmits would invalidate the cached tensors.**
> > I think, since you now added quite some logic around to_tensor, it would be nice to add a separate test for it, so we know what behavior is expected, and it will not break in the future. Seems there is not a lot of testing for to_tensor. Especially now the complexity of to_tensor has grown with the multiple arguments and flexibility it supports.
> > As you hint to yourself, we probably also want to have a look how to make to_tensor more consistent across probe, scan and config. But maybe in a later PR :).
> > **BTW; this bug might be solved in #23. if the dependency resolution was broken and you rely on it with to_tensor, this could fix it already.**
> 
> to verify, this bug was not introduced in this PR right? I noticed it on the main branch as well, see #28.

I believe this is a different bug. On the main branch, the test as well as the notebook seems to run. This bug is related to caching of tensors as far as I can see.

wesselvannierop (dismissed):

zea/internal/parameters.py
tristan-deep: I think compute and skip missing can be removed. If you are including a tensor, then you also want it to be computed (otherwise omit it from the include list). Skip missing i would not recommend, better for the user to clearly specify what parameters it wants. I think that would be less error prone.

zea/ops.py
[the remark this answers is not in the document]
wesselvannierop: Let's convert this to an issue.

zea/internal/parameters.py
```diff
@@ -308,6 +317,7 @@ def _invalidate(self, key):
         self._cache.pop(key, None)
         self._computed.discard(key)
         self._dependency_versions.pop(key, None)
+        self._tensor_cache.pop(key, None)
```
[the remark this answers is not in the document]
wesselvannierop: This happens but will cleanup

zea/internal/parameters.py
```diff
@@ -349,40 +359,70 @@ def _resolve_dependency_tree(self, name, failed=None):
...
-            tensor_dict[key] = _to_tensor(key, val)
+                else:
+                    raise exc
+
+            if key in self._tensor_cache:
+                tensor_dict[key] = self._tensor_cache[key]
+            else:
+                tensor_val = _to_tensor(key, val)
+                tensor_dict[key] = tensor_val
+                self._tensor_cache[key] = tensor_val
```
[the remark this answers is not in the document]
wesselvannierop: _invalidate will handle that

zea/internal/parameters.py
```diff
@@ -349,40 +358,70 @@ def _resolve_dependency_tree(self, name, failed=None):
...
+            keys = keys - set(exclude)
+
+        tensor_dict = {}
+        # Convert parameters and computed properties to tensors
+        for key in keys:
+            # Get the value from params or computed properties
+            try:
+                val = getattr(self, key)
+            except MissingDependencyError as exc:
+                if include == "all":
```
tristan-deep, line 413: @wesselvannierop do we want to add a case for when a user uses exclude (i.e. implicitly setting "all" parameters except the one you define)?

tristan-deep:
Also, during the merge of the notebooks into this PR, was able to _drastically_ reduce file size (10MB to 500KB) of the zea [sequence example notebook](https://github.com/tue-bmd/zea/blob/feature/parameter_tensor_caching/docs/source/notebooks/pipeline/zea_pipeline_example.ipynb). Turns out that some small tweaks to how you render the animation go a long way. Probably good to keep this in mind with future notebooks as well.

tristan-deep (approved):

PR #23, opened 2025-06-30
Clean-up and improve selecting transmits

wesselvannierop:
- Automatically discovers which attributes depend on `selected_transmits` (and this found some missing ones from the hardcoded list).
- `scan.selected_transmits = [0]` calls `scan.set_transmits`, before this gave a non-descriptive recursion error.
- [Fix for using set_transmits with a numpy array](https://github.com/tue-bmd/zea/pull/23/commits/c4cfbd2cfd79cd73d40f52e7b503c72228fecbe6)

### DIFF
--- a/tests/test_scan.py
+++ b/tests/test_scan.py
@@ -109,6 +109,22 @@ def test_selected_transmits_affects_shape(attr, expected_shape):
     else:
         assert val.shape == (expected_shape[0],)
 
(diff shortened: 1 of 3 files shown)


tristan-deep (approved):
Maybe add a test for the bug you initially fixed as well? Otherwise looks good!

PR #25, opened 2025-07-02
Add option for polar grid to the Scan class

wesselvannierop:
Mainly implements the selection of a polar grid in the scan class ([and wrote a test for it](https://github.com/tue-bmd/zea/blob/23b9d39f795291f01f609dd783074cb7a43a9b15/tests/test_transmit_schemes.py#L495-L540)). Fixes some other issues along the way. 
Also wrote a notebook:

[Open In Colab](https://colab.research.google.com/github/tue-bmd/zea/blob/feature/polar_grid/docs/source/notebooks/pipeline/polar_grid.ipynb) [View on GitHub](https://github.com/tue-bmd/zea/blob/feature/polar_grid/docs/source/notebooks/pipeline/polar_grid.ipynb)

Other fixes:

- Computing `coordinates` in the `zea.Scan` class for `zea.ops.ScanConvert`.
- Fixes `zea.Pipeline.append`, `prepend()` and adds `insert()`. Previously, if your pipeline was `with_batch_dim=True` and you were adding an `Operation` that was the opposite, the Pipeline broke. Now it reinitializes the Pipeline.
- Fixes #28

PR #29, opened 2025-07-03
Adds `zea.Folder.copy` method

wesselvannierop:
Merge after #25

This PR adds the `zea.Folder.copy` method which is useful to copy a folder of `zea.File`'s to a new location, e.g., an AI compute cluster. By default only copies one key (and the metadata such as scan object), this is useful to keep the dataset size down. Can also copy all keys. Will only copy what is needed, e.g., if the key is already present in the destination it will just skip it.


tristan-deep (changes requested):
Nice, good feature! I tested with:

```python
import zea

path = "hf://zeahub/picmus"
to_path = "test_output_folder"

folder = zea.Folder(path, key="raw_data")

folder.copy(to_path=to_path, all_keys=True)
```

Few things I noticed:
- You need to specify a key for folder. This was already the case before this update, but maybe we can change it here? What if you just want to inspect a folder, and don't know what keys it has, this would be a bit annoying. Some properties depend on the key however, so in that case a default key or something would be necessary. 
- The folder structure is not preserved in this case. I would opt to keep the folder structure by default, and flatten as an option?
- It now prints copying dataset from, but does not show where it copies to.

wesselvannierop:
> * It now prints copying dataset from, but does not show where it copies to.

https://github.com/tue-bmd/zea/blob/9770d54ea8c913ad13bce4f1ac219633edc76cf8/zea/data/datasets.py#L395-L399

It does print where it copies to right?

tristan-deep:
> > * It now prints copying dataset from, but does not show where it copies to.
> 
> https://github.com/tue-bmd/zea/blob/9770d54ea8c913ad13bce4f1ac219633edc76cf8/zea/data/datasets.py#L385-L387
> 
> It does print where it copies to right?

You are right! Not sure why I didn't notice it at first, maybe it printed on next line. Works for me now I check!

zea/data/__main__.py
```diff
@@ -0,0 +1,31 @@
...
+Usage:
+    python -m zea.data <source_folder> <destination_folder> <key>
+"""
+
+import argparse
+
+from zea import Folder
+
+
+def main():
```
tristan-deep, line 12: Just seeing this now, but seems like a handy CLI. Maybe document it on the data page so people know it exists?

tristan-deep (approved):

Issue #24, opened 2025-07-01
docker image are not properly tagged on dockerhub


tristan-deep:
Currently docker images are not properly tagged, see any of the images [here](https://hub.docker.com/u/zeahub). Just see: latest, cached, and all the SHAs. Besides properly tagging upon release (see [publish.yaml](https://github.com/tue-bmd/zea/blob/main/.github/workflows/publish.yaml) for an example), we could also clean up the intermediate builds.

tristan-deep:
Fixed in #33. For example see the tag [here](https://hub.docker.com/r/zeahub/all). You can do:

```shell
docker pull zeahub/all:v0.0.2
```

Something a bit related. I wonder if we still need dockerhub, since we can also host the packages on github, see [here](https://github.com/tue-bmd/zea/pkgs/container/zea%2Fall-cpu). We already do this for our CI images, but deploy images are still on dockerhub.

PR #33, opened 2025-07-08
Refactor CI/CD Pipeline

tristan-deep:
> [!NOTE]
> Diffs were checked manually since code was moved across new files, but please since important feel free to check as well. Basic logic remains almost unchanged, except for the `set-tag` fix, see #24.

## Changes Made

- Split the monolithic CI pipeline into separate, focused workflows
- Created reusable components to eliminate code duplication
- Removed conditional logic from workflows - they only run when needed
- Production Docker images are no longer built on every main push (was slow and redundant)
- Fixed Docker tagging bug - images now get proper version tags instead of just SHAs (see #24)

## New Workflow Structure

| Workflow | Triggers | Purpose |
|----------|----------|---------|
| **Tests** | PRs + `main` pushes | Unit tests, heavy tests, test docs build |
| **Linter** | PRs + `main` pushes | Code formatting and linting |
| **Docker Images** | `docker/*` branches + any tag | Production Docker images |
| **PyPI Publish** | GitHub releases | Build and publish Python package |

## Reusable Components

- `set-tag.yaml`: Centralized Docker tag logic
- `build-image.yaml`: CI image with dev dependencies and all backends (CPU)

## Workflow Execution

**Pull Request / Push to main:**
```
[set-tag → build-image → [tests + heavy-tests + docs-build] (parallel)] + [linter] (parallel)
```

**Push to `docker/*` branch or tag:**
```
set-tag → build-production-images (all backends, CPU and GPU)
```

**GitHub Release:**
```
build → publish-to-pypi
```


tristan-deep:
@wesselvannierop now ready for review. 

---

Something I was thinking of for future changes, we could add a fast parameter to notebooks tests. And then decrease number of steps or iterations in certain cells dependent on that fast parameter. I see that there are 3 notebooks that are problematic (see example of a run [here](https://github.com/tue-bmd/zea/actions/runs/16153735631/job/45591703353)):
- custom_models_example.ipynb in 698.0s
- diffusion_model_example.ipynb in 475.1s
- zea_sequence_example.ipynb in 845.4s

The other example notebooks are all < 100s.

Alternatively, we can run the heavy tests in a GPU environment (they are run on self-hosted anyways). That would also speed this up a lot. On the other hand it would need to build the GPU image, which seems to be a bit slow (also with caching weirdly enough?).

wesselvannierop (approved):

PR #35, opened 2025-07-09
Changed warning that implied pytorch is not supported

Louisvh:
When running the examples, _jit_compile is hit and issues a warning that torch is not supported. It didn't specify it's talking about JITting though, so it looked like the package was telling the user torch isn't supported altogether.

### DIFF
--- a/zea/backend/__init__.py
+++ b/zea/backend/__init__.py
@@ -1,4 +1,4 @@
-"""Backend subpackage for ``zea``.
+"""Backend-specific utilities.
 
 This subpackage provides backend-specific utilities for the ``zea`` library. Most backend logic is handled by Keras 3, but a few features require custom wrappers to ensure compatibility and performance across JAX, TensorFlow, and PyTorch.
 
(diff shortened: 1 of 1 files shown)


tristan-deep:
Btw, for the tests to run, let's wait for #33 to merge. That will fix the problems.

zea/backend/__init__.py
```diff
@@ -108,7 +108,7 @@ def _jit_compile(func, jax=True, tensorflow=True, **kwargs):
         return func
     else:
         log.warning(
-            f"Unsupported backend: {backend}. Supported backends are 'tensorflow' and 'jax'."
+            f"JIT compilation not currently supported for backend {backend}. Supported backends are 'tensorflow' and 'jax'."
```
tristan-deep, line 111: Just for reference, would the example work if you init the Pipeline with jit_compile=False. Might be also nice to add a line, "if you are using a zea.Pipeline, try to initialize with jit_compile=False. Just to hint the user to a solution if they still want to use the torch backend.

tristan-deep (approved):

tristan-deep (approved):
Nice!

PR #36, opened 2025-07-09
Fix cache when switching backend

wesselvannierop:
When you would run the exact same code after switching to a different backend, our cache mechanism would break (as found by @Louisvh). This adds the keras backend to the hash. Also fixes a potential bug in the `Parameter._current_dependency_hash` that I introduced :)


zea/internal/cache.py
```diff
@@ -120,7 +120,10 @@ def serialize_elements(key_elements: list):
             element = hashlib.md5(element).hexdigest()
             serialized_elements.append(element)
 
-    return serialized_elements
+    serialized = "_".join(serialized_elements)
+    if shorten:
```
tristan-deep, line 124: Any particular reason for adding this?

tristan-deep (approved):
Looks good!

PR #49, opened 2025-07-14
Fixed jit_options being ignored for the PatchedGrid in Pipeline.from_default()

Louisvh:
When running the default pipeline with torch and `jit_options=None`, there would still be a warning that jit is not supported and it's falling back to the non-jitted version because the PatchedGrid didn't see the parent's kwargs. The Pipeline kwargs are now also passed to the `PatchedGrid` call in `from_default()`, which correctly sets the `jit_options` and skips the jit attempt altogether.


Louisvh:
This is a bit messy because https://github.com/tue-bmd/zea/pull/35 hadn't been merged yet, so the "files changed" list includes those changes. The diff from the tip of that PR to this one is only [this one-line change](https://github.com/tue-bmd/zea/pull/49/commits/9c1ccbe43f639fdbbbfdd4177c42b05211846e67).

Issue #53, opened 2025-07-14
pydicom bug


tristan-deep:
I think I broke versioning in https://github.com/tue-bmd/zea/commit/15b9a99edf1b9ec43b64601e610274e12fd24d60 by prematurely updating the pydicom api, which breaks [`pydicom.pixels.convert_color_space`](https://pydicom.github.io/pydicom/stable/reference/generated/pydicom.pixels.convert_color_space.html). Either we should update pydicom in our toml (>3.0 I believe) or remove pydicom altogether. We don't really support dicom throughout anyways.

tristan-deep:
Agreed!

PR #54, opened 2025-07-15
Removed pydicom dependency

Louisvh:
Pydicom was broken (see: issue #53); this PR removes it altogether. We only used it for loading dicom videos in the iolib, but it wasn't actually checking the video metadata properly before loading, so the current implementation would actually fail for most dcm videos. We could at some point re-add support, but then we should probably support all versions of dicom video, rather than just one of the options; maybe better to ask users to convert any dicom videos they want to use into a more predictable format first.


tristan-deep (approved):
Perfect! 🚀

PR #61, opened 2025-07-16
Speed up slow notebook test

Louisvh:
There was a 25GB download lurking around in one of the notebook tests. I've replaced it with a 2.5GB version that only contains the data actually used in the example.


tristan-deep:
> I think this might be nice for changing of parameters just for the tests: https://papermill.readthedocs.io/en/latest/usage-parameterize.html?highlight=parameters

Ah that is perfect!

PR #63, opened 2025-07-17
Prevent error when probe is unknown

vincentvdschaft:
Using `zea.data.load_file` with a file with an unknown probe name would give an error because the probe_geometry in the file would be compared to a `None` value. This PR fixes this by only doing this check if both `probe_geometry` values are not `None`.


wesselvannierop:
Although the PR on itself is probably okay, I wonder if this logic should anyhow be improved. I suggest we initialize a probe object based on the parameters that are in the data file, and use the probe name only to find values that are not in the data file perhaps?

tristan-deep:
Okay issue was made here #66. @vincentvdschaft could you add a test for the line you added?

tristan-deep:
@wesselvannierop I now simplified the probe from file logic. It will always return you a generic probe, instantiated with the parameters in the file. I think this is the behavior we want anyways. Only downside is that it looses the name that is stored in the file (that is not transferred to the probe class). I think you would generally only do `Probe.from_name`, in case you don't have a data file, for instance when you simulate data.

wesselvannierop:
We could of course still check if the probe geometry is the same with the definition in the code, based on the name. Not sure if that is really needed though
```python
if self.probe_name in probe_registry:
...
```

tristan-deep:
> We could of course still check if the probe geometry is the same with the definition in the code, based on the name. Not sure if that is really needed though
> 
> ```python
> if self.probe_name in probe_registry:
> ...
> ```

Yes was thinking of that solution as well. You would still need to overwrite the parameters from the file in that Probe from name object. But would not be too bad of a solution. That would also fix the failing test i think, and also the logic would be cleaner compared to how it is now.

wesselvannierop:
Might not even have to overwrite. Just check if 
```python
Probe.from_name().probe_geometry == probe_parameters_file["probe_geometry"]
```

tristan-deep:
> Might not even have to overwrite. Just check if
> 
> ```python
> Probe.from_name().probe_geometry == probe_parameters_file["probe_geometry"]
> ```

Should maybe also take into account other probe parameters.

tristan-deep:
@wesselvannierop lemme know what you think of current implementation.

PR #67, opened 2025-07-21
Greedy entropy bug fix

OisinNolan:
* Discovered a bug in `GreedyEntropy` for the case where `n_possible_actions` =/= `img_width`. 
* This was caused by `ops.reshape` stacking vectors in the wrong order.
* The fix involves adding a transpose before and after the reshape to make sure that the order is correct.
* I also added a regression test that just asserts the line selected by greedy entropy with a hard coded 'correct answer', which should help prevent any unintended regressions if we change things in future.
* Also an extra minor change I made it so that `entropy_sigma` can be specified in the constructor rather than being hard coded.

Edit by @wesselvannierop:
* Added missing transpose to `CovarianceSamplingLines`, and add regression test for this too.

### DIFF
--- a/zea/agent/selection.py
+++ b/zea/agent/selection.py
@@ -92,6 +92,7 @@ class GreedyEntropy(LinesActionModel):
         mean: float = 0,
         std_dev: float = 1,
         num_lines_to_update: int = 5,
+        entropy_sigma: float = 1.0,
     ):
         """Initialize the GreedyEntropy action selection model.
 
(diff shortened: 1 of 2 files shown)


zea/agent/selection.py
```diff
@@ -92,6 +92,7 @@ def __init__(
         mean: float = 0,
         std_dev: float = 1,
         num_lines_to_update: int = 5,
+        entropy_sigma: float = 1.0,
```
wesselvannierop, line 95: Maybe you can add a line to the docstring what `entropy_sigma` does?

wesselvannierop (approved):

PR #65, opened 2025-07-21
Clean up and refactor

wesselvannierop:
Changes

- Improves documentation (mainly of `Scan` class)
- `Nx` and `Nz` --> `n_x` and `n_z`
- Remove global static parameters and add to ops.
- Refactor `Parameters` class into clean and reusable functions
- Fix `pipeline.prepend` issues with batch dims
- Add `ops.ChannelsToComplex` and `ops.ComplexToChannels`
- `rx_apo` works with `PatchedGrid`
- `vfocus` --> `focus_distances`

TODO:

- [ ] Rename `Nx` and `Nz` in huggingface to `n_x` and `n_z`

PR #68, opened 2025-07-22
Sync configs to HF

tristan-deep:
Syncs the [configs directory](https://github.com/tue-bmd/zea/tree/main/configs) to the configs folder on [zeahub HF](https://huggingface.co/datasets/zeahub/configs). See discussion in #65, [here](https://github.com/tue-bmd/zea/pull/65#issuecomment-3095850946).


tristan-deep:
One problem I'm seeing now is that notebook tests might still fail when you change a config, since it will grab the old config from HF (which only gets updated after we merge a PR)...

tristan-deep:
> The `test_all_configs_valid_hf` should be removed now right?

Yes, since it will test "outdated" configs. Since configs in repo are now synced, and they are already tested, the test is also redundant.

PR #74, opened 2025-08-04
Fix error when only one type of waveform

vincentvdschaft:
The `generate_zea_dataset` function would raise an exception when only one of `waveforms_two_way` and `waveforms_one_way` was supplied.


tristan-deep (approved):
lgtm!

PR #75, opened 2025-08-05
Diffusion call choose network and training type

tristan-deep:
This PR allows for a bit more flexibility during runtime of the diffusion model to choose between training / inference mode and choosing ema or main network during the call. Default behavior should remain the same (where ema network was always chosen during inference and not during training). You can now also run main network with training=False, which can be nice for instance if you sample from a diffusion model during training and want to backprop through the main network. 

## Summary by CodeRabbit

* **New Features**
  * Added options to explicitly select the network variant ("main" or "ema") during diffusion and denoising processes.
  * Introduced the ability to disable JIT compilation in the reverse diffusion process.
  * Enhanced control over training and inference modes within diffusion workflows.

* **Refactor**
  * Improved code clarity in the diffusion schedule calculation without changing its behavior.


wesselvannierop (approved):
lgtm

PR #79, opened 2025-08-15
Make all unary functions from keras.ops available as zea.Operation

wesselvannierop:


wesselvannierop:
Discussion points:

1. Do we want everything to be available under `zea.ops`?
    - How to handle name conflicts?
2. Maybe everything under `zea.keras.ops` or `zea.keras` to create clear seperation?
3. What if we want to extent the functionality of the basic `keras.ops`, such as now with `zea.ops.Stack` or `zea.ops.Mean`?
4. Maybe we want the functional API under `zea.ops.functional`? (related to #82)

Issue #20, opened 2025-06-25
Scan conversion issue with TF backend


tristan-deep:
Scan conversion with TF backend seems to be broken. A quick fix is to set `order=2` (which defers to scipy under the hood, that is why it works), or change the Keras backend to "jax" or "torch". TF will return the full array with only fill value for every element (for instance zeros or nan). Seems to be a TF issue of the underlying [map coordinates func](https://github.com/keras-team/keras/blob/v3.10.0/keras/src/ops/image.py#L741) (see maybe related info [here](https://github.com/tensorflow/tensorflow/issues/34788) and [here](https://github.com/keras-team/keras/issues/20608)).

### Quick example to reproduce

```python
import os

os.environ["KERAS_BACKEND"] = "tensorflow"

import keras
import matplotlib.pyplot as plt
import numpy as np

from zea.display import scan_convert_2d

n_rho, n_theta = 128, 128
polar_img = np.zeros((n_rho, n_theta), dtype=np.float32)

arc_rho = np.arange(n_rho)
arc_theta = n_theta // 4
polar_img[arc_rho, arc_theta] = 1.0
polar_img[arc_rho, n_theta - arc_theta - 1] = 1.0

# Define polar coordinate ranges
rho_range = (0, 100)
theta_range = (-np.pi / 4, np.pi / 4)  # -45 to +45 degrees

# Scan convert
polar_img = keras.ops.convert_to_numpy(polar_img)
cartesian_img, params = scan_convert_2d(polar_img, rho_range=rho_range, theta_range=theta_range)
cartesian_img = keras.ops.convert_to_numpy(cartesian_img)

# Plot
fig, axs = plt.subplots(1, 2, figsize=(10, 4))
axs[0].imshow(polar_img, aspect="auto", cmap="gray")
axs[0].set_title("Polar Image")
axs[0].set_xlabel("Theta index")
axs[0].set_ylabel("Rho index")

axs[1].imshow(cartesian_img, aspect="auto", cmap="gray")
axs[1].set_title("Scan Converted (Cartesian)")
axs[1].set_xlabel("X index")
axs[1].set_ylabel("Z index")

plt.tight_layout()
plt.show()
```

![Image](https://github.com/user-attachments/assets/f81a369a-79e0-4f2a-9248-52792025dc9f)

But should give this (which is the case for the other backends, e.g. try `os.environ["KERAS_BACKEND"] = "jax"`)

[...]

PR #89, opened 2025-08-29
Fix tensorflow map coordinates (Keras 3.11)

tristan-deep:
See #20 

## Summary by CodeRabbit

- New Features
  - Improved TensorFlow/Keras support for scan conversion, aligning behavior with other backends.

- Chores
  - Raised minimum required Keras version to 3.11.

- Refactor
  - Streamlined backend handling and unified numeric conversions to reduce backend-specific checks.

- Tests
  - Expanded cross-backend test coverage to include TensorFlow; simplified test descriptions and adjusted tolerances and dtypes for consistent results.


tristan-deep:
> For completeness: I tested if I could increase the decimal number in the tests for tensorflow. But for high resolution scan conversion, there are apparently slight deviations, that are not visible for me, but will still fail the test if decimal>0.

Yes I had the same conclusion, thanks for checking!

PR #121, opened 2025-10-10
Fix Scan pickling and equality comparison after pickling

### LINKED ISSUE #110: Allow to dump a scan instance to a file
Currently when using pickle to dump / load a scan instance, we get a recursion error:

```python
import pickle

with open('scan.pkl', 'rb') as f:
    scan = pickle.load(f)
Traceback (most recent call last):
  File "<string>", line 5, in <module>
  File "/ulsa/zea/zea/internal/parameters.py", line 264, in __getattr__
    if item in self._params:
               ^^^^^^^^^^^^
  File "/ulsa/zea/zea/internal/parameters.py", line 264, in __getattr__
    if item in self._params:
               ^^^^^^^^^^^^
```

wesselvannierop:
Fixes #110, adds test for scan pickle and improves `serialize_elements`

## Summary by CodeRabbit

* **New Features**
  * Pipeline factory gains an optional timing flag to enable lightweight timing instrumentation.

* **Refactor**
  * Unified internal serialization and hashing used for cache keys and object checksums, improving consistency and reliability while preserving public behavior.

* **Tests**
  * Added a round-trip serialization test for Scan objects to confirm serialize/deserialize produce an equal but distinct instance.
  * Updated a performance test to use vectorized computation and adjusted timing expectations.

### DIFF
--- a/zea/data/file.py
+++ b/zea/data/file.py
@@ -219,6 +219,27 @@ class File(h5py.File):
     def load_data(self, data_type, indices: str | int | List[int] = "all"):
         """Load data from the file.
 
+        The indices parameter can be used to load a subset of the data. This can be
+
+        - 'all' to load all data
+
+        - an int to load a single frame
+
+        - a list of ints to load specific frames
+
+        - a tuple of lists, ranges or slices to index frames and transmits. Note that
+          indexing with lists of indices for both axes is not supported. In that case,
+          try to define one of the axes with a slice.
+
+        .. code-block:: python
+
+            # Load frame 5
+            File.load_data("raw_data", indices=5)
+            # Load frames 0, 2 and 4
+            File.load_data("raw_data", indices=[0, 2, 4])
+            # Load frames 0-9 and transmits 0, 2 and 4
+            File.load_data("raw_data", indices=(slice(10), [0, 2, 4]))
+
         Args:
             data_type (str): The type of data to load. Options are 'raw_data', 'aligned_data',
                 'beamformed_data', 'envelope_data', 'image' and 'image_sc'.


wesselvannierop:
I think I agree. However there are instances in `zea` where `shorten=False`, for example: `Parameters.serialized` which is used for comparing scan instances with each other. These serializations can be hashed and it will still work, but not sure how big the risk of a hash collision is?

wesselvannierop:
I did keep the `serialize_elements` because it can be nice on its own but moved the hashing to the function `hash_elements` which is basically what previously was `shorten=True`.

tristan-deep:
> I did keep the `serialize_elements` because it can be nice on its own but moved the hashing to the function `hash_elements` which is basically what previously was `shorten=True`.

I got a review comment on that function (that it was buggy) in #126. Was gonna look into it but since you changed maybe not necessary anymore? See https://github.com/tue-bmd/zea/pull/126#discussion_r2429043591

wesselvannierop:
> > I did keep the `serialize_elements` because it can be nice on its own but moved the hashing to the function `hash_elements` which is basically what previously was `shorten=True`.
> 
> I got a review comment on that function (that it was buggy) in #126. Was gonna look into it but since you changed maybe not necessary anymore? See [#126 (comment)](https://github.com/tue-bmd/zea/pull/126#discussion_r2429043591)

The comment you linked is another function

tests/test_utils.py
```diff
@@ -189,28 +189,32 @@ def test_block_until_ready_timing():
...
+        # Vectorized heavy computation (no Python for-loop):
+        # - Build an outer-product matrix (n x n)
+        # - Apply elementwise trig operations
+        # - Reduce to a scalar
+        y = ops.matmul(x[:, None], x[None, :])  # shape (n, n)
+        z = ops.sin(y) + ops.cos(y)
+        return ops.sum(z, axis=1)
 
-    x = ops.ones(1000)
+    x = ops.ones(10_000)
```
tristan-deep: nice

Issue #50, opened 2025-07-14
Offset error in beamforming


vincentvdschaft:
I think we have an error in the beamformers. We define t=0 as the moment the first element starts firing and then compute the travel time to a gridpoint and back. What we do not consider here is that the actual peak of the waveform is not when the element starts firing, but some time later. This means that we beamform to the wrong part of the signal leading in a shift in location and incorrect focussing. The figure below illustrates what I mean:

<img width="1279" height="655" alt="Image" src="https://github.com/user-attachments/assets/2017d487-4fe8-4628-801f-7a1bf431ee0a" />

I think the solution is to have the offset of the peak as an additional input to the beamformers.

tristan-deep:
Fixed in #131

PR #125, opened 2025-10-14
Adding waveforms and TGC to scan, changing waveform saving

vincentvdschaft:
# Changes
- The waveform and time gain compensation (TGC) curve have been an optional item in the data format for some time, but are not included in the `Scan` object. I now added them. Having the waveforms available will enable resolving #50.
- In doing this I realized that the way the waveforms are stored in the data format makes this more difficult. The way the waveforms are stored is as `waveforms_two_way\waveform_000`, `waveforms_two_way\waveform_001`. This nested shape was designed like this because waveforms might have different lengths. I now think it is actually nicer to save the waveforms as a single array and just pad the shorter ones. I think the wasted space is negligible and it does simplify the structure. Let me know if you agree with this.
- To be compatible with files with the nested structure for now, I added a conversion function with a deprecation warning. This warning will only show up for files that have the waveforms stored in them.

# How to test
- Convert a raw data file to zea format.
- Run a pipeline to beamform to check if it works.
- Verify that the waveforms and TGC are present in the scan object.

## Summary by CodeRabbit

* **New Features**
  * Added support for TGC gain curves, one‑way/two‑way waveforms, and per‑transmit waveform indices with accessors and sensible defaults when values are missing.

* **Refactor**
  * Waveforms are stored and written as aggregated arrays; legacy dict-style waveform inputs are auto-converted with deprecation warnings and empty/None inputs normalized.

* **Tests**
  * Expanded tests to validate shapes and presence of the new waveform, index, and TGC parameters.

### DIFF
--- a/zea/scan.py
+++ b/zea/scan.py
@@ -220,8 +220,9 @@ class Scan(Parameters):
     }
 
     def __init__(self, **kwargs):
-        # Store the current selection state before initialization
-        selected_transmits_input = kwargs.pop("selected_transmits", None)
+        # Ensure that selected_transmits is present and set to None by default
+        selected_transmits_input = kwargs.get("selected_transmits", None)
+        kwargs["selected_transmits"] = None
 
         # Initialize parent class
         super().__init__(**kwargs)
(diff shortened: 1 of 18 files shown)


zea/scan.py
```diff
@@ -536,6 +547,39 @@ def time_to_next_transmit(self):
...
+            return None
+        return value
+
+    @property
+    def waveforms_two_way(self):
+        """The two-way transmit waveforms of shape (n_waveforms, n_samples)."""
+        value = self._params.get("waveforms_two_way")
+        if value is None:
+            return None
+        return value
```
wesselvannierop: These properties do not change the default behavior and can be removed, `__getattr__` will handle them.

vincentvdschaft: Ah alright. I removed them.

Issue #127, opened 2025-10-14
Incorrect missing dependency error with selected_transmits


vincentvdschaft:
The automatic caching feature in the `Scan` erroneously states that there is a missing dependency when a property is accessed that is not in the scan object.

# To reproduce
```python
from zea.data.data_format import generate_example_dataset
from zea.data.file import load_file

path = "zea.hdf5"
# Create an example dataset
generate_example_dataset(path)

data, scan, probe = load_file(path)
print(scan.selected_transmits)
print(scan.focus_distances)
```

This code first prints the selected transmits, proving that they are present and then generates this error:
```bash
zea/zea/internal/parameters.py", line 397, in _assert_dependencies_met
    raise MissingDependencyError(name, missing_set)
zea.internal.parameters.MissingDependencyError: Cannot access 'focus_distances' due to missing dependencies: ['selected_transmits']
```

wesselvannierop:
Fixed in #128

PR #128, opened 2025-10-14
Fix dependency error with selected_transmits in Scan object

vincentvdschaft:
This PR fixes the following issue
- #127 

The scan object has the `selected_transmits` parameter, that other parameters depend on, but it would be kept out of the `Parameters` update logic to set it manually. This would result in `selected_transmits` not being in `self._params` and would cause the error from #127 when a parameter that is dependent on `selected_transmits` is accessed.

# Changes
- The `Scan` object now also passes the `selected_transmits` to the super class
- When setting the `selected_transmits` the `set_transmits` function is still called, but this is now also passed to the super class.

# How to test
- Please refer to the snippet in the issue

## Summary by CodeRabbit

* **Bug Fixes**
  * Scans now default safely when no transmit selection is provided and correctly apply any provided selection.
  * Selection updates reliably persist and reflect the resolved state, avoiding silent no-ops.
  * More robust and simplified update logic reduces cases where selections might not appear updated.

* **Tests**
  * Added a test ensuring accessing valid-but-unset scan attributes does not raise errors.

### DIFF
--- a/tests/test_ops_infra.py
+++ b/tests/test_ops_infra.py
@@ -222,6 +222,19 @@ def patched_pipeline():
     return pipeline
 
 
+def test_pipeline_modification():
+    """Tests if modifying the pipeline updates callable layers correctly."""
+    # set timed to True to ensure _callable_layers is used
+    # basically this makes sure that the pipeline is reinitialized
+    pipeline = ops.Pipeline.from_default(jit_options=None, with_batch_dim=False, timed=True)
+    pipeline.prepend(ops.Simulate())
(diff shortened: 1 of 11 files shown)


tristan-deep:
Nice! Clean fix, can we add your issue as a test maybe? To catch that bug in the future.

wesselvannierop:
The fix looks good! I took the liberty to remove some redundant lines I discovered during reviewing. Like, @tristan-deep suggested, a test case for the bug would be good.

wesselvannierop (approved):

PR #139, opened 2025-10-23
Manually specify output keys of Pipeline for more robust tracing

### LINKED ISSUE #129: Pipeline needs_keys needs attention
Using `ast` to find the return keys of `Operation` is fragile and fails when the return statement does not adhere to return {...}

https://github.com/tue-bmd/zea/blob/d37de6bf91011e2f336a36f174cb5667d97b63bf/zea/ops.py#L205-L235

See some problematic cases:
Image

This was introduced in #117 to fix the case when an operation needs a key that a previous operation already provides. The reason we need to know which keys the pipeline needs up front is because of:

https://github.com/tue-bmd/zea/blob/d37de6bf91011e2f336a36f174cb5667d97b63bf/zea/ops.py#L1044-L1048

wesselvannierop:
Fixes #129 by making the operations specify which output keys they will produce.

Additionally improves the test for `block_until_ready`.

## Summary by CodeRabbit

* **New Features**
  * Operations now support explicit tracking of multiple output keys per operation, enabling advanced data routing and flexible branching within pipelines.
  * Pipelines provide enhanced visibility into all output keys produced across operations.

* **Improvements**
  * Parameter processing now gracefully handles non-existing keys in include lists, improving robustness during configuration.


tristan-deep (approved):
Nice! Did we have a test already for this special case where a required parameter of the pipeline is generated in an intermediate stage of the pipeline itself (but not passed at the input)?

Issue #130, opened 2025-10-15
Improve dataset_info.yaml


wesselvannierop:
There are currently some issues with the `"dataset_info.yaml"` file which is generated by the dataloader(s).

1. The file it produces has these:
```yaml
file_shapes:
- !!python/tuple
  - 124
  - 500
  - 686
```
so we need to convert the `tuple`s to `list`s before dumping to yaml.

2. The `hdf5_key_for_length` arg to `search_file_tree` determines which key is used to store the file shapes. However, not all keys will have the same file shape (i.e. `image_sc.shape != image.shape`)! This needs to be fixed.

3. The `Folder` has a optional argument `key` but it seems this is actually needed for the code to run.
https://github.com/tue-bmd/zea/blob/d37de6bf91011e2f336a36f174cb5667d97b63bf/zea/data/datasets.py#L167-L175

You can avoid these bugs by passing `redo=True` and `write=False` to `search_file_tree` or equivalently using `search_file_tree_kwargs` to `Folder` and `Dataset`.

Also, we might want to support read-only folders, where we skip writing to a `"dataset_info.yaml"`.

swpenninga:
> Just checking, was point 3 also fixed in [#130](https://github.com/tue-bmd/zea/issues/130) [@swpenninga](https://github.com/swpenninga) ? I can open a separate issue just for that one if it isn't.

Yeah I do not know what to do about this. if the hdf5 key is `None` the code will still crash like before.
I do not know if there are some datasets out there that use different filetypes (.png, .jpeg, etc.) and still function though, maybe a separate issue to look into this is a good idea.

wesselvannierop:
Point 3 is also fixed now in #161

PR #161, opened 2025-11-03
Fix search_file_tree regarding issue #130

swpenninga:
Refactors the logic for getting files in a dataset folder and fetching the shapes. Now uses the already existing `cache_output` logic instead of a `dataset_info.yaml`. You can enable / disable parallelism in the `find_h5_files` using the env variable `ZEA_FIND_H5_SHAPES_PARALLEL`.

If the dataset changes, will automatically invalidate the cached shapes. All of this does not require write permissions on the dataset folder, because it uses `ZEA_CACHE_DIR`.

## Summary by CodeRabbit

* **New Features**
  * Simplified file discovery now yields matching file paths directly.
  * Faster HDF5 shape discovery with caching and optional parallel processing.

* **Bug Fixes**
  * Validation/metadata write failures now log warnings instead of raising errors.
  * More robust handling when shape data for a key is missing.

* **Documentation**
  * Added env var to control parallel HDF5 shape retrieval.

* **Tests**
  * Tests configured to disable parallel shape retrieval by default.


swpenninga:
Again, I suggest that after merging all `dataset_info.yaml` files are deleted as they become outdated. 
They can still be used, but they will contain artifacts from the old convention.

wesselvannierop:
@swpenninga sorry for hijacking your PR... I was working on an implementation to invalidate `dataset_info.yaml` when the dataset content changes, but I thought a refactor would be in place, so now the logic is quite different, but much simpler. Let me know what you think!

Issue #151, opened 2025-10-28
Set fixed seed to prevent spurious failing of test due to precision


tristan-deep:
@tristan-deep I think the tests that are failing should be fixed by setting a fixed seed for the inputs. Also, there are many more tests that have `np.random` inside the `pytest.mark.parametrize` so we might want to do a single PR that fixes all those.

_Originally posted by @wesselvannierop in https://github.com/tue-bmd/zea/issues/150#issuecomment-3455547778_

PR #165, opened 2025-11-04
Standardize RNG in all tests

swpenninga:
This PR is to ensure that tests will not fail anymore due to `np.random`, as stated in issue #151. 
RNG is now standardized with minimal changes to the tests:

1. Previously, data was generated/selected in three ways; 

- `np.random` without seed
- `np.random` with seed
- `rng.random` generator from `np.random.default_rng(seed)`

This is now all standardized to the latter.

2. The seeds were either non-present or hardcoded into a generator, with values `2`, `42`, `123` or `12345`. Now, every file that contains any RNG has a global variable `DEFAULT_SEED = 42` at the top, which any RNG function will use.

3. I explicitly made sure `rng = np.random.default_rng(DEFAULT_SEED)` is not a global variable, since this will introduce variability if tests are ran in a different order, or tests are added later. If RNG is used, the generator will be created in the test locally.

4. There are exceptions to the rules above, like in [test_agent.py](https://github.com/tue-bmd/zea/blob/main/tests/test_agent.py) where the tests outcome is hardcoded to work with seed 2. 
I left things this way, because making the seed a variable breaks the test if the value ever changes. I added a comment about this in the code.

## Summary by CodeRabbit

* **Tests**
  * Introduced a shared default test seed and applied it across the test suite for consistent RNG initialization.
  * Test data generation now consistently uses seeded RNGs (including switches to normally distributed generators where appropriate) for deterministic, reproducible inputs.
  * Test helpers now disable timeouts when a debugger is attached to aid debugging.
  * No test logic or public behavior changed; only randomness and test-time behavior were unified.

### DIFF
--- a/tests/__init__.py
+++ b/tests/__init__.py
@@ -21,3 +21,5 @@ run_in_backend = backend_workers.run_in_backend
 DUMMY_DATASET_N_FRAMES = 4
 DUMMY_DATASET_GRID_SIZE_Z = 256
 DUMMY_DATASET_GRID_SIZE_X = 256
+
+DEFAULT_TEST_SEED = 42
--- a/tests/helpers.py
+++ b/tests/helpers.py
@@ -14,7 +14,6 @@ import jax
 import numpy as np
 import pytest
 
-DEFAULT_SEED = 42
 
 debugging = sys.gettrace() or debugpy.is_client_connected() is not None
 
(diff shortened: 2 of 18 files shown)


tristan-deep (changes requested):
Nice! Would only change the global variable `DEFAULT_SEED` location to the test init file so it is only defined once for all tests. 

EDIT: i would also rename to `DEFAULT_TEST_SEED` or something with test, to communicate to developers this is for testing only.

tristan-deep (approved):
looks good now!

PR #168, opened 2025-11-06
Make PatchedGrid 3D-compatible

vincentvdschaft:
`PatchedGrid` currently does not work with 3D pixelgrids. This PR fixes this by removing the explicit 2D reshape.

# How to test
This is some example code that results in an error on main, but does run after this PR.
```python
import os

import keras
import numpy as np

from zea.data import load_file
from zea.ops import (
    Pipeline,
)

path = "hf://zeahub/picmus/database/experiments/contrast_speckle/contrast_speckle_expe_dataset_iq/contrast_speckle_expe_dataset_iq.hdf5"

data, scan, probe = load_file(
    path=path,
    indices=[0],
    data_type="raw_data",
)

# index the first frame
data_frame = data[0]

scan.n_ch = 2  # IQ data, should be stored in file but isn't currently
scan.xlims = probe.xlims
scan.zlims = (0, 0.06)  # reduce z-limits a bit for better visualization
dynamic_range = (-50, 0)  # set dynamic range for display
pipeline = Pipeline.from_default(
    num_patches=100,
    baseband=True,
    pfield=False,
    with_batch_dim=False,
    jit_options="pipeline",
)

parameters = pipeline.prepare_parameters(probe, scan)
parameters.pop("dynamic_range", None)  # remove dynamic_range since we will set it manually later

x_grid, y_grid, z_grid = np.meshgrid(
    np.linspace(scan.xlims[0], scan.xlims[1], 32),
    np.linspace(scan.xlims[0], scan.xlims[1], 128),
    np.linspace(scan.zlims[0], scan.zlims[1], 64),
    indexing="ij",
)
grid = keras.ops.array(np.stack((x_grid, y_grid, z_grid), axis=-1))
flatgrid = keras.ops.reshape(grid, (-1, 3))

print(grid.shape)


parameters["grid"] = grid
parameters["flatgrid"] = flatgrid
inputs = {pipeline.key: data_frame}

# dynamic parameters can be freely passed here as keyword arguments
outputs = pipeline(**inputs, **parameters)

image = outputs[pipeline.output_key]
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Patch-based processing now derives patch shapes from actual input grid dimensions, improving accuracy for variable grid sizes.
  * Grid/patch context is consistently threaded through processing calls, preventing missing or incorrect patch inputs and stabilizing outputs.

* **Improvements**
  * Expanded input handling lets callers provide explicit grid/patch context, increasing flexibility and reliability of patch operations.


tristan-deep:
Do you think we can also adapt PatchedGrid to work with for instance Demodulate?

wesselvannierop (dismissed):

tristan-deep:
> @tristan-deep There is for sure ways to avoid OOMs for demodulate, but there is no `grid` in `Demodulate` and `PatchedGrid` loops over the grid. So let's tackle that in another PR.

Yeah you are right, this would be more a batch in sequence / map op. Would also need to be less specialized than `PatchedGrid`.

wesselvannierop (dismissed):

Issue #143, opened 2025-10-23
Redesign beamformer blocks


tristan-deep:
Currently it is unclear whether the beamformer is an ops or a pipeline and this can be improved quite a bit. I propose to split up the beamformer in more basic ops, and then compose the beamformer as a pipeline. Since pipelines are easily nestable this then can be integrated in the B-mode pipeline. Furthermore it allows us to specify different types of beamformers such as DAS, MV or ABLE. 

Something like this. We would then remove the `DelayAndSum` op, which now does not do delay anyways (`TOFCorrection` does this).

```python
beamformer = PatchedGrid(
    operations=[
        TOFCorrection(),
        Reshape(),
        AdaptiveWeighting(),
        Sum(),
    ]
) # is a Pipeline itself now
```

wesselvannierop:
```python
import zea


beamformer = zea.Pipeline([
    zea.ops.PatchedGrid([
        TOFCorrection(),
        AdaptiveWeighting(),
        Sum(),
        Sum(),
    ]),
    Reshape(),
])
```

tristan-deep:
> > [@vincentvdschaft](https://github.com/vincentvdschaft) given the current implementation would you still consider this or is the current f-number setting sufficient?
> 
> The reason I mentioned it is because I implemented some functionality for smooth window functions for the f-number windowing which are now not available when using the pipeline (see link) 
> - #19
> 
> I thought it might be more flexible to have the f-number as a separate operation to allow people to use the different windows that are there or implement their own. I am not sure if it is necessary though. If you think this would be useful I can have a go at it. What do you think?
> 
> 
> 

I think so, would be quite useful!

tristan-deep:
@vincentvdschaft I added `zea.ops.ReceiveApodization` in #482. You can set `f_number` to zero and apply the receive apodization operation on your own. We [also discussed](https://github.com/tue-bmd/zea/pull/482#pullrequestreview-4663793533) to have a better integration / workflow, that will have to be future work. Closing this now, feel free to reopen or create a PR in case you have a good idea or disagree with closing this.

PR #176, opened 2025-11-20
Generalize `zea.ops.PatchedGrid` to `zea.ops.Map`

wesselvannierop:
Introducing `zea.ops.Map` -- this is a type of Pipeline, that maps its operations over specified input arguments. This generalizes `PatchedGrid`, which I have kept that for backwards compatibility, but it is implement as a subclass of `zea.ops.Map`.

You can now specify any keys (together with `in_axes` and `out_axes`, like in jax.vmap) to map over, not just `flatgrid` and `flat_pfield`!

To do this I had to make the reshaping of the grid a separate operation `zea.ops.ReshapeGrid`. So to run your old Pipelines, you now have to append that after `PatchedGrid`! This is one of the steps towards #143.

Additionally, this PR will `log.debug` whenever you give the pipeline keys that it will not use, useful if you think some operation is using them but is actually not.

## Summary by CodeRabbit

* **New Features**
  * Flexible batching operator with explicit controls for input/output axes, chunking and batch sizing.
  * New pipeline reshape step available in reconstruction/beamforming flows.
  * One-time warnings for unexpected pipeline inputs.

* **Improvements**
  * Simplified beamforming workflows by removing implicit reshaping and exposing batching controls.
  * Pipeline configs and notebook outputs now surface batching and logging details for easier tuning.

* **Documentation**
  * Corrected grid dimension naming in docs.

### DIFF
--- a/zea/data/preset_utils.py
+++ b/zea/data/preset_utils.py
@@ -1,6 +1,6 @@
 """Preset utils for zea datasets hosted on Hugging Face.
 
-See https://huggingface.co/zea/
+See https://huggingface.co/zeahub/
 """
 
 from pathlib import Path
--- a/zea/models/__init__.py
+++ b/zea/models/__init__.py
@@ -72,7 +72,7 @@ The following steps are recommended when adding a new model:
 
 1. Create a new module in the :mod:`zea.models` package for your model: ``zea.models.mymodel``.
 2. Add a model class that inherits from :class:`zea.models.base.Model`. For generative models, inherit from :class:`zea.models.generative.GenerativeModel` or :class:`zea.models.deepgenerative.DeepGenerativeModel` as appropriate. Make sure you implement the :meth:`call` method.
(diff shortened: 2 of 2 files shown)


zea/ops.py
```diff
@@ -1256,31 +1278,72 @@ def pipeline_to_yaml(pipeline: Pipeline, file_path: str) -> None:
...
+        operations: List[Operation],
+        argnames: list,
+        in_axes=0,
+        out_axes=0,
+        chunks=None,
+        batch_size=None,
+        **kwargs,
+    ):
+        super().__init__(operations, **kwargs)
+        self.argnames = argnames
```
benluijten: For my understanding:
Does Map chunk all tensors/keys that are given by argnames? And if so, does this mean that for all tensors/keys the same axes are used? 

I'm thinking of the scenerio where you want to map over e.g. transmits, which is typically axis index 1 ([n_batch, n_tx, ...]. And apodizations which might be shaped [n_tx, n_el], aka the transmits axis is index 0.

tristan-deep (changes requested):
see comment above, requires some clarification on usage in the docs.

zea/ops.py
wesselvannierop: This you can use `in_axes` for, like in `jax.vmap`

wesselvannierop:
@benluijten @tristan-deep I've added a snippet and improved the documentation. Let me know if you think this is sufficient!

PR #198, opened 2025-12-05
Added tests for dataset conversion to zea format

swpenninga:
Regarding #194 and #173 

Added tests for 
- `echonet.py`, `camus.py`, `picmus.py` (marked heavy)
- `images.py`, `utils.py`

Other small changes:
1. SimpleITK had to be added to the test runners to be able to test CAMUS. It is not added as a zea dependency.
2. Exception for SimpleITK in the CAMUS conversion script to tell a user to manually install it.
3. test_dataloader now inherits global variable

Did not write a test for echonetlvh yet since it still needs changes #182.
Did not write a test for verasonics because I do not understand the code well enough to test it.
They have placeholder functions that are already formatted like the rest.

## Summary by CodeRabbit

* **Tests**
  * Added comprehensive dataset conversion tests (conversion flows, unzip, AVI/image handling) and switched tests to use the project-wide test seed.
* **Dependencies**
  * Added SimpleITK to test requirements.
* **Improvements**
  * Better, actionable error message when SimpleITK is missing.
* **Chores**
  * CI test workflows updated to install test extras and include coverage uploads.

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>

PR #200, opened 2025-12-08
Clean-up workflow files

tristan-deep:
There are some redundant lines in our workflow files that can be cleaned up.

## Summary by CodeRabbit

* **Chores**
  * Simplified CI by removing injected user and environment flags from container runs and deleting the standalone heavy-test step for more consistent test execution.
  * Slimmed the runtime image by removing multi-user and sudo setup, reducing runtime complexity.

* **Documentation**
  * Updated installation/run docs to remove user-specific Docker run guidance, recommend rootless Docker, and simplify examples.

### DIFF
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -122,9 +122,6 @@ jobs:
       - name: Checkout repository
         uses: actions/checkout@v4
 
-      - name: Set up Git
-        run: git config --global safe.directory $GITHUB_WORKSPACE
-
       - name: Run heavy Pytest
         run: |
           docker run --rm \


tristan-deep:
This should also fix the codecov report for heavy tests which were not registered properly. See https://github.com/tue-bmd/zea/pull/198#issuecomment-3631462347

wesselvannierop (approved):
lgtm

.github/workflows/tests.yaml
```diff
@@ -118,19 +114,6 @@ jobs:
           token: ${{ secrets.CODECOV_TOKEN }}
           slug: ${{ env.MAIN_REPO }}
 
-      - name: Run heavy Pytest (allowed to fail)
```
OisinNolan: Out of curiosity was there a particular reason for this to be in both `tests` and `heavy-tests` or was it just accidentally duplicated?

PR #211, opened 2025-12-15
Improve verasonics conversion script

wesselvannierop:
## Fixes
- [x] Fixes `focus_distances`. Verasonics stores this in wavelengths, we assume it is in meters. This caused some ultrasound reconstruction being suboptimal. Users need to fix their previously converted data files, but we also have a detection mechanism that will auto-convert.
- [x] Add `allow_accumulate`. Sometimes, some transmits are already accumulated
                on the Verasonics system (e.g. harmonic imaging through pulse inversion).
                In this case, the mode in the Receive structure is set to 1 (accumulate).
                If this flag is set to False, an error is raised when such a mode is detected.
- [x] Newly saved Verasonics workspaces [using `vera.saving.save_raw`](https://github.com/tue-bmd/veratools/blob/main/%2Bvera/%2Bsaving/save_raw.m), did not reorder probe elements. The python conversion script handles this now.
- [x] Improve CLI (and docs) such that the specific args are only shown per dataset type
- [x] Order of frames in `data/image` and `data/raw_data` was not right, because of the circular buffer used by Verasonics. The first frame of the sequence is often somewhere in the middle of the buffer. This is fixed now.
- [x] Run on cpu only. Meaning the operations inside the conversion do not initialize a gpu.
- [x] Add test for conversion script (conversion CLI changed a bit!)

## Gotcha's
- The recorded frames in `data/raw_data` and `data/image` doesn't seem to match. Generally, the `image` buffer does not contain the same frames as the raw data buffer. This happens because the Verasonics often does not reconstruct every acquired frame. This means that the images often skip frames, and span a longer time period than the raw data buffer.

## Summary by CodeRabbit

* **New Features**
  * Added Verasonics dataset conversion workflow, modular dataset-specific CLI, new scan properties for waveform handling, and exposed log_compress in public API.

* **Bug Fixes**
  * Normalized focus-distance units, improved HuggingFace path resolution and safer file read behavior.

* **Documentation**
  * Updated beamforming parameter docs.

* **Tests**
  * Expanded conversion test coverage for Verasonics, improved test data generation and subprocess output capture.

### DIFF
--- a/tests/test_pfield.py
+++ b/tests/test_pfield.py
@@ -3,6 +3,7 @@
 import numpy as np
 
 from zea.beamform.delays import compute_t0_delays_planewave
+from zea.ops import Pipeline
 from zea.probes import Verasonics_l11_4v
 from zea.scan import Scan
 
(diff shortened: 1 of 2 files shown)


zea/ops/ultrasound.py
```diff
@@ -97,12 +97,7 @@ class TOFCorrection(Operation):
     """Time-of-flight correction operation for ultrasound data."""
 
     # Define operation-specific static parameters
-    STATIC_PARAMS = [
-        "f_number",
-        "apply_lens_correction",
-        "grid_size_x",
-        "grid_size_z",
-    ]
+    STATIC_PARAMS = ["f_number", "apply_lens_correction"]
```
vincentvdschaft: This seems unrelated to the rest of this PR. That is fine of course, but please confirm this should be in here.

vincentvdschaft (dismissed):
Looks nice and works with both 2d and 3d data. I placed two comments but other than that I think it is good to go.

zea/ops/ultrasound.py
wesselvannierop: Good one, yes I found that these args are no longer used so I removed them here.

PR #214, opened 2025-12-18
Fix `PatchedGrid` computing pfield even when not needed

### LINKED ISSUE #203: `PatchedGrid` requires pfield computation whether or not pfield weighting is used
The `PatchedGrid` operation tries to map over the `flat_pfield` property of the Scan class. This triggers pfield computation, even if the `PfieldWeighting` operation is not in the pipeline.

A workaround is the simply us the `Map` function, i.e.:

Replace:
```python
PatchedGrid([TOFCorrection(), DelayAndSum()], chunks=128),
```
With:
```python
Map([TOFCorrection(), DelayAndSum()], argnames="flatgrid", chunks=128),
```

In general it goes a bit against the principle of having modular operations for an operation to assume that some other operation is present in the pipeline. One solution is to deprecate `PatchedGrid` and stick with `Map`, and just use good error messages to help people know which `argnames` they should specify.

wesselvannierop:
Closes #203 

## Summary by CodeRabbit

* **Refactor**
  * Simplified Map pipeline key validation by removing its custom overrides and relying on the base pipeline behavior.
* **Tests**
  * Added a test ensuring pfield computation is skipped when pfield support is disabled in pipeline configuration.

### DIFF
--- a/tests/test_pfield.py
+++ b/tests/test_pfield.py
@@ -3,6 +3,7 @@
 import numpy as np
 
 from zea.beamform.delays import compute_t0_delays_planewave
+from zea.ops import Pipeline
 from zea.probes import Verasonics_l11_4v
 from zea.scan import Scan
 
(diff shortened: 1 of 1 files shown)


tristan-deep:
Let's add a test for this as well

PR #219, opened 2025-12-22
Enable HDF5 compression for zea data format

vincentvdschaft:
This PR adds an option to enable compression to `generate_zea_dataset`. This can reduce the file size by a lot at the cost of slower saving and loading.

In my test with a relatively large data file I get these times and sizes:
|file size|time to run convert (cli)|
|-|-|
|1.5GB | 120s|
|5.4GB | 39s|

In case you think this is a good addition we should think about adding a CLI flag to `zea.data.convert` to enable/disable compression.

## Summary by CodeRabbit

* **New Features**
  * Datasets now use gzip compression by default for non-scalar data, reducing storage size.
  * Users can disable compression when creating datasets; the conversion tools include a command-line option to turn off compression.

### DIFF
--- a/zea/data/convert/verasonics.py
+++ b/zea/data/convert/verasonics.py
@@ -966,7 +966,14 @@ class VerasonicsFile(h5py.File):
 
         return frame_indices
 
-    def to_zea(self, output_path, additional_functions=None, frames="all", allow_accumulate=False):
+    def to_zea(
+        self,
+        output_path,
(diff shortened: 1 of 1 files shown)


wesselvannierop (approved):
Seems like a great addition! I took the liberty to add the `--no_compression` arg to the CLI. I wonder if data loading will get much slower from a compressed file but I think it is worth the immense space savings

Issue #194, opened 2025-12-04
Tests for zea conversion scripts missing.


swpenninga:
Not all scripts in zea/data/convert have a test

- [x] echonet
- [x] echonetlvh
- [x] camus
- [x] picmus
- [x] verasonics
- [x] images
- [x] utils

Issue #155, opened 2025-10-30
Black areas in Echonet LVH images


wesselvannierop:
The 'un scan converted' images in echonetlvh sometimes show large black padding around the images. This might indicate an issue in `cartesian_to_polar_matrix`.

For example:
`echonetlvh/train/0X1027077445AE5512.hdf5`

`image_sc`:

<img width="804" height="762" alt="Image" src="https://github.com/user-attachments/assets/f3ed9b05-d7e2-4297-a1cc-55f631c60a23" />

`image`:

<img width="804" height="762" alt="Image" src="https://github.com/user-attachments/assets/8ad3f913-6f8b-438b-a2fd-45fb2bc3c047" />

Additionally, it seems the max value is `254` instead of `255` in the images:

```python
file = zea.File("/data/USBMD_datasets/echonetlvh/train/0X1027077445AE5512.hdf5")

image_sc = file.load_data("image_sc", indices=0)
image_sc.max() # => 254 instead of 255 ??

image = file.load_data("image", indices=0)
image.max() # => 254 instead of 255 ??
```

PR #220, opened 2025-12-30
EchoNetLVH Conversion Tests + Fixes

OisinNolan:
**Summary**
- This PR addresses two current issues related to EchoNetLVH dataset conversion:
    1) #194 --> adds testing for the conversion script
    2) #155 --> fixes bug that would add additional black areas into the polar images. The issue here was that the scan cone angle detected by `fit_and_crop_around_scan_cone` wasn't being passed to the cartesian -> polar conversion [here](https://github.com/tue-bmd/zea/blob/47c5795dc28175ab80fb4c8cdfd660a3f98a319d/zea/data/convert/echonetlvh/__init__.py#L284), so it would assume that the scan cone angle is pi radians. The true angle was often <pi radians, meaning that additional surrounding black pixels were being included. I also updated `cartesian_to_polar_matrix` to allow the scan cone apex to be between two pixels, which is the case for images with an even width.
- Once these changes are finalized + merged we can regenerate EchoNetLVH with these fixes

## Release Notes

* **New Features**
  * Added option to specify a custom rejection file path when converting datasets.
  * Opening angle parameter now included in cone parameter data output.

* **Bug Fixes**
  * Improved precision in coordinate transformations and image cropping calculations for more accurate results.

* **Tests**
  * Expanded test coverage for EchoNet-LVH dataset handling.

### DIFF
--- a/tests/data/test_conversion_scripts.py
+++ b/tests/data/test_conversion_scripts.py
@@ -190,7 +190,6 @@ def create_echonetlvh_test_data(src):
         src (Path): path to the source directory where test data will be created.
     """
     extra_args = []
-    extra_args.append("--no_hyperthreading")
 
     from zea.display import scan_convert_2d
 


tristan-deep (approved):
Good to have this! And seems like an extensive test with some of the edge cases covered as well. Maybe @swpenninga can look at it a bit as well given that he created the testing framework for the conversion scripts. But lgtm!

PR #241, opened 2026-01-22
Add batch_dim to simulate

swpenninga:
Previously `simulate_rf` had an output shape of
``` 
(1, n_tx, n_ax, n_el, 1)
```
where the first dim was a batch dimension that was appended.

This does not work with the zea `Operations` format, as it uses `ops.map()` which adds another batch dim.

This only worked because batch simulations were not implemented, and all tests happened to have a batch_size of 1, meaning they passed while the code was actually not working. 

After implementing batched simulations, the whole pipeline breaks down `(test_default_ultrasound_pipeline in test_ops_infra)`, so I edited the simulate_rf output shape to
```
(n_tx, n_ax, n_el, 1)
```

Now everything is fixed, and i made a test for both batched and non-batched behaviour.

## Summary by CodeRabbit

* **New Features**
  * Added optional batch-dimension handling across phantom generation and simulation inputs.

* **Bug Fixes**
  * Normalized simulation output layout by removing an unexpected leading singleton and aligning shapes with batched/unbatched inputs.

* **Tests**
  * Tests parameterized for batched and unbatched scenarios; inputs and assertions updated for new shapes.

* **Documentation**
  * Example notebook and minor test formatting updated to reflect batched scatterer shapes.

### DIFF
--- a/docs/source/notebooks/metrics/lpips_example.ipynb
+++ b/docs/source/notebooks/metrics/lpips_example.ipynb
@@ -120,7 +120,7 @@
     {
      "data": {
       "application/vnd.jupyter.widget-view+json": {
-       "model_id": "f667b2a0780b4cb09a36906617012652",
+       "model_id": "6cfd915761fc4f55bea8f3cc82ccf982",
        "version_major": 2,
        "version_minor": 0
       },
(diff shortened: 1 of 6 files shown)


tristan-deep (approved):
nice! 🌽

Issue #255, opened 2026-02-09
HDF5 dataloaders using a python generator


swpenninga:
Any dataloaders that use the `H5Generator` are effectively singlethreaded and cpu-bound:

https://github.com/tue-bmd/zea/blob/2a3a9a62755af26e30769357b5e95a60df3ed393/zea/data/dataloader.py#L376

Either I am completely misunderstanding the situation, but as far as I know:

```
tf.data.Dataset.from_generator(image_extractor, ...)
```

Means that __next__() runs in Python. This gives the pipeline: 

```
GPU waits → Python → h5py → Python → TF → GPU
```

and `num_workers`, `AUTOTUNE`, `prefetch`, and `batch`  have no effect in speeding this up.

Currently, I'm training an RF-data VAE that's taking 7s to load a batch, and 120ms to do the forward and backward pass which is why I cannot use the zea dataloader.

---

I understand that fixing this is quite a large task as there are many dependencies, but it would be good to look into at some point.

tristan-deep:
Working on the Grain solution in #256. From initial testing I think it would be good to transition to the Grain dataloader. 

We would have to create some requirements / benchmarks so we actually build the dataloader that suits our needs.

PR #256, opened 2026-02-09
Move dataloader to Grain

tristan-deep:
_[Grain](https://google-grain.readthedocs.io/en/stable/index.html) is a library for reading data for training and evaluating JAX models. It’s open source, fast and deterministic._

- [x] Implement grain dataloader
- [x] [Benchmark dataloaders](https://google-grain.readthedocs.io/en/latest/tutorials/dataset_debugging_tutorial.html) for common dataloading scenarios
- [x] Investigate chunking hdf5 @OisinNolan 
- [x] Check if all functionality is preserved w.r.t. the old dataloader
- [x] Add tests for new dataloader / adjust existing tests
- [x] Remove old dataloader

I was able to implement most of the functionality of the existing dataloader, making it a drop in replacement for `zea.backend.tensorflow.make_dataloader`. Simply use `zea.Dataloader` instead. 

I would like to move the dataloader to grain (from Tensorflow previously). There are a few advantages:
- This will remove the heavy Tensorflow dependency in case you would like to use the dataloader with a different backend
- Simple, works with numpy operations
- Works for all backends
- More modern dataloader
- With the limited testing I did, it was faster (see benchmark todo)
- Hopefully resolves issue we found in #255 

```python
import zea

dataset_path = "hf://zeahub/camus-sample/val"
dataloader = zea.Dataloader(
    dataset_path,
    key="data/image_sc",
    batch_size=4,
    shuffle=True,
    clip_image_range=[-60, 0],
    image_range=[-60, 0],
    normalization_range=[0, 1],
    image_size=(256, 256),
    resize_type="resize",  # or "center_crop or "random_crop"
    num_threads=16,
    seed=4,
)

for batch in dataloader:
    print("Batch shape:", batch.shape)
    break  # Just show the first batch

fig, _ = zea.visualize.plot_image_grid(batch)
fig.savefig("test.png")

```

## Summary by CodeRabbit

* **New Features**
  * Added a framework-agnostic Dataloader for HDF5 datasets, exposed at the package top level.

* **Breaking Changes**
  * Removed the legacy TensorFlow-specific dataloader and the old H5Generator-style API.
  * Removed the legacy I/O retry utility and file-info JSON helpers.

* **Documentation**
  * Updated example notebooks and docs to use the new Dataloader API.

* **Chores**
  * Added Grain >= 0.2 to project dependencies.

* **Tests**
  * Updated and expanded tests to cover the new dataloader and data-source behaviors.

### DIFF
--- a/zea/data/dataloader.py
+++ b/zea/data/dataloader.py
@@ -299,7 +299,7 @@ class H5DataSource:
 
         if self.return_filename:
             file_data = {
-                "fullpath": file.path,
+                "fullpath": file.filename,  # same as file.path, but str type
                 "filename": file.stem,
                 "indices": indices,
             }


tristan-deep:
> Currently working on the new grain backend. @tristan-deep I understand the dataloader is fully np based, however I was wondering about the current `_numpy_translate()` and `_numpy_resize()`.
> 
> These are expensive operations, is it not better to do the bare minimum I/O on cpu -> off-load to GPU _(working on prefetching)_ -> do these transformations with keras ops?

That would make a lot of sense actually if that is possible in an efficient way! I did notice indeed the operations you mention take a lot of time in the current implementation.

PR #309, opened 2026-04-13
Fix echonet shape tracing with JAX backend

tristan-deep:
`EchoNetDynamic.build(input_shape)` is called with the shape of the user input. Inside `maybe_convert_to_jax`, the TF SavedModel was traced using that same shape. However, `self.network` is called in `call()` *after* the image has already been resized to (1, 112, 112, 3) and tiled to 3 channels.  The SavedModel only has concrete functions for shape (None, 112, 112, 3), so an error is raised. Fixed now by fixing the shape to INFERENCE_SIZE.

Snippet to reproduce bug (fixed after this PR)

```python
import os

os.environ["KERAS_BACKEND"] = "jax"  # only JAX backend is affected by this bug

import numpy as np

import zea
from zea.models.echonet import EchoNetDynamic

zea.init_device()

model = EchoNetDynamic.from_preset("echonet-dynamic")

# Simulate a full-resolution grayscale frame (different from 112x112)
image = np.random.rand(552, 669).astype("float32")

# Before the fix, the build step traced the SavedModel at (1, 552, 669, 1),
# which has no matching concrete function, causing a ValueError.
# After the fix, tracing uses the correct (1, 112, 112, 3) shape.
masks = model(image[None, ..., None])

print("SUCCESS — masks shape:", masks.shape)

```

## Summary by CodeRabbit

* **Refactor**
  * Simplified model conversion to use a fixed example input shape and relaxed shape checks during converted-model execution for more robust behavior.

* **Style**
  * Improved formatting and clarity of an import-failure message.


OisinNolan (approved):
nice fix, lgtm

PR #310, opened 2026-04-13
Add overlay_masks func

tristan-deep:
Added overlay mask function to enhance getting started snippet. Also that snippet now saves the image, so it is clear to a starter what the output should be.

> [!NOTE]  
> Note that one should also merge the bugfix in #309 to run the snippet. 

## Summary by CodeRabbit

* **New Features**
  * Added an image mask-overlay utility with adjustable transparency, color support, broad input-format handling, and argument validation.

* **Documentation**
  * Updated Getting Started with a streamlined data-loading/preprocessing/inference example, a different example model/preset (with an onnxruntime note), and an expanded mask-visualization walkthrough including 8‑bit conversion and display.

* **Tests**
  * Added unit tests covering overlay behavior across image/mask types, modes, non-PIL inputs, and alpha settings.


docs/source/getting-started.rst
```diff
@@ -35,26 +35,41 @@ Let's take a quick look at how to use ``zea`` to load and process ultrasound dat
...
+   masks = keras.ops.squeeze(masks)
+
+   image = zea.display.to_8bit(image)
+
+   masks_clipped = keras.ops.where(masks > 0.5, 255, 0)
+
+   masks_clipped = zea.display.to_8bit(masks_clipped, dynamic_range=(0, 255))
 
+   result = zea.display.overlay_masks(image, [masks_clipped], alpha=0.5)
+   result.save("result.png")
```
swpenninga: Maybe more of a design choice, but I find it a bit odd that the getting-started has an example codeblock that saves a result that is then not shown. Other than that, the PR itself is a nice addition!

swpenninga (dismissed):

docs/source/getting-started.rst
tristan-deep: ah good point, maybe i can do `result.show()` then? I'm too used to headless mode...

swpenninga:
Is it ready for re-review?

swpenninga:
I guess the change from `image.save()` to `image.show()` does not do anything in the docs? (outside of the notebooks which are saved in an executed state) 

https://zea--310.org.readthedocs.build/en/310/getting-started.html

If the intent of this Getting started example is for people to do it themselves, then this is fine i guess.

swpenninga (approved):

PR #349, opened 2026-05-06
Improve device handling

### LINKED ISSUE #27: Better support for multi GPU
Currently, one can select a GPU as follows:

```python
import zea
device = zea.init_device("auto:1")
```

Depending on the backend however, this will not enforce for instance a pipeline to run on that device. Would be nice if we can pass the device to init of a pipeline, or to prepare parameters, to actually put the tensors on the correct device.

tristan-deep:
Improved device handling with `zea.device` (previously `zea.backend.on_device`). Relies on [`keras.device`](https://github.com/keras-team/keras/blob/6a83d3936aaa81e5c1798a4d71f2b3928d8fdc52/keras/src/backend/__init__.py#L75). 

`zea.init_device` now can also work with selecting multiple GPU devices (#27) and will return a list. e.g.

```python
devices = zea.init_device("auto:2")
print(f"Selected devices: {devices}")
# >>> ["gpu:0", "gpu:1"]
```

Now also easy to run a pipeline on a specific GPU. Either with context manager:
```python
pipeline = zea.Pipeline([zea.ops.keras_ops.Abs()])
with zea.device("gpu:1"):
    output = pipeline(data=data)["data"]
```

or argument to pipeline:
```python
zea.Pipeline([zea.ops.keras_ops.Abs()], device="gpu:1")
```

A full example:

```python
import os

os.environ["KERAS_BACKEND"] = "torch"  # or "jax" or "torch"

import zea

# Select the 2 GPUs with auto:2, which selects the ones with most free memory; hides the rest.
# Returns ["gpu:0", "gpu:1"] for TF/JAX, ["cuda:0", "cuda:1"] for torch.
devices = zea.init_device("auto:2")

import keras
import numpy as np

print(f"Selected devices: {devices}")

data_np = np.array([-3.0, -1.0], dtype=np.float32)

pipe0 = zea.Pipeline([zea.ops.keras_ops.Abs()], device=devices[0])
pipe1 = zea.Pipeline([zea.ops.keras_ops.Abs()], device=devices[1])

inp = keras.ops.convert_to_tensor(data_np)
print(f"Input tensor device: {inp.device}")

out0 = pipe0(data=inp)["data"]
out1 = pipe1(data=inp)["data"]

print(f"pipe0 ({devices[0]}) → pipeline: {out0.device}, tensor: {out0.device})")
print(f"pipe1 ({devices[1]}) → pipeline: {out1.device}, tensor: {out1.device})")

# or with context managers:
pipe0 = zea.Pipeline([zea.ops.keras_ops.Abs()])
with zea.device(devices[0]):
    t0 = pipe0(data=inp)["data"]
    print(f"pipe0 context manager → pipeline: {pipe0.device}, tensor: {t0.device})")

pipe1 = zea.Pipeline([zea.ops.keras_ops.Abs()])
with zea.device(devices[1]):
    t1 = pipe1(data=inp)["data"]
    print(f"pipe1 context manager → pipeline: {pipe1.device}, tensor: {t1.device})")

```

## TODO
 - [x] Add option to select GPUs for running pytest locally
 - [ ] Make sure zea.init_device is used correctly in all examples (also notebooks)

## Summary by CodeRabbit

* **New Features**
  * New top-level device context for explicit device placement.
  * Pipeline supports instance-level and per-call device placement and includes device in serialized params.

* **Improvements**
  * Deferred ML backend imports for faster, lighter package import.
  * More consistent multi‑GPU selection and CPU fallback behavior across backends.

* **Tests**
  * GPU-marked tests auto-skip when no CUDA GPU is available; expanded device/device-placement test coverage.

* **Documentation**
  * Documentation index updated to list the device entry.

[Review Change Stack](https://app.coderabbit.ai/change-stack/tue-bmd/zea/pull/349?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


wesselvannierop:
There is still the classic problem that `zea` imports `keras` which imports `torch` and after `torch` has been imported changing `CUDA_VISIBLE_DEVICES` doesn't work. We could return the actual device numbers for torch instead of hiding the GPUs. But I feel like that is error prone on shared GPU servers. Alternatively, we have to find a way to keep `zea` from import `keras` before `init_device` is called.

wesselvannierop:
I made claude swap to lazy imports for zea, now it works correctly (if you really start with `init_device` before importing anything else). We can have a look if this is something we want @tristan-deep ?

tristan-deep:
> I made claude swap to lazy imports for zea, now it works correctly (if you really start with `init_device` before importing anything else). We can have a look if this is something we want @tristan-deep ?

Okay let's go for this. Probably good to also mention the behavior of zea.init_device with and without hiding devices in the module docstring. So people know where to use it. I guess we also need to change our example scripts (which call zea.init_device too late, at least for torch).

wesselvannierop:
I still have some (previously hidden) circular imports to fix.

tests/tools/test_hf.py
```diff
@@ -88,8 +88,8 @@ def fake_list_files(repo_id, repo_type="dataset", **kwargs):
         assert repo_type == "dataset"
         return fake_files
 
-    monkeypatch.setattr("zea.data.preset_utils._hf_parse_path", fake_parse_path)
-    monkeypatch.setattr("zea.data.preset_utils._hf_list_files", fake_list_files)
+    monkeypatch.setattr("zea.internal.preset_utils._hf_parse_path", fake_parse_path)
+    monkeypatch.setattr("zea.internal.preset_utils._hf_list_files", fake_list_files)
```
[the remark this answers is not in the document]
wesselvannierop: Solved now?

tests/__init__.py
```diff
@@ -6,9 +6,7 @@
 DEFAULT_TEST_BACKEND = "tensorflow"
 os.environ["KERAS_BACKEND"] = DEFAULT_TEST_BACKEND
 os.environ["TF_CPP_MIN_LOG_LEVEL"] = "1"
-os.environ["CUDA_VISIBLE_DEVICES"] = ""
```
wesselvannierop, line 9: How do running tests on a shared machine set the device?

tristan-deep: Hmm I removed this because with pytest and --gpu flag we can now test some gpu required tests (regarding device setting). But I see the problem... Maybe we can add init device here too in the argparse?

wesselvannierop: That might work indeed

wesselvannierop:
The notebook tests seem to suddenly take a long time to run. I think it might be related to huggingface, but will see if this also happens in another PR.

Edit: seems like other PRs have the same issue (https://github.com/tue-bmd/zea/actions/runs/25853067702/job/75964346878)

tristan-deep:
> The notebook tests seem to suddenly take a long time to run. I think it might be related to huggingface, but will see if this also happens in another PR.
> 
> Edit: seems like other PRs have the same issue (https://github.com/tue-bmd/zea/actions/runs/25853067702/job/75964346878)

But resolved now again?

wesselvannierop (dismissed):

wesselvannierop (dismissed):

docs/source/notebooks/models/custom_models_example.ipynb
```diff
@@ -81,6 +81,7 @@
     "import zea\n",
     "from zea import load_file, init_device\n",
     "from zea.visualize import set_mpl_style\n",
+    "from zea import backend\n",
```
tristan-deep:
```python
import zea

zea.device
```

I believe `zea.device` is in the main namespace.

docs/source/notebooks/models/custom_models_example.ipynb
```diff
@@ -227,15 +229,16 @@
     }
    ],
    "source": [
-    "parameters = pipeline.prepare_parameters(probe, scan, dynamic_range=dynamic_range)\n",
+    "parameters = pipeline.prepare_parameters(probe, scan, dynamic_range=dynamic_range, device=device)\n",
```
tristan-deep, line 231: Should we remove device here? Since you already use the context manager below.

wesselvannierop: The tensors are not moved with the context manager. But we could put prepare_parameters inside the context manager here?

tristan-deep: Ah true, only if initialized within the scope, i guess. In that case the context manager doesn't do much in this specific instance? I'd say leave like this then.

tristan-deep: Also what you suggest is probably cleanest, preparing them on the correct device immediately.

zea/ops/pipeline.py
```diff
@@ -1112,21 +1115,13 @@ def get_dict(self, compact=True) -> dict:
         if not compact or self.enable_pfield:
             params["enable_pfield"] = self.enable_pfield
 
-        # Pipeline-level params
```
tristan-deep, line 1115: just checking, was this intended to be removed?

wesselvannierop: These now come in through super().get_dict()

docs/source/notebooks/models/custom_models_example.ipynb
wesselvannierop: Sure, but note they are already prepared on the right device with the device=device arg.

wesselvannierop (approved):
Good to go for me

PR #352, opened 2026-05-14
Dataloader: option to disable re-shuffle

wesselvannierop:
Now you can set `reshuffle_each_epoch=False` for the dataloader, such that you do get a random shuffle, but it doesn't change every time you call `iter(dataloader)`.

Additionally sneaked in:
- Fixes a bug in the gumbel softmax code while running in the `torch` backend.
- Also fixes a bug I encountered which is that the dataloader used `keras.ops`, which may not be set to the numpy backend.

## Summary by CodeRabbit

* **New Features**
  * Optional per-epoch reshuffling via constructor option.
  * Public shuffle(seed) to rebuild/reseed the pipeline on demand.
  * Optional final conversion-to-tensor and explicit conversion-to-numpy after resizing.

* **Refactor**
  * Image preprocessing moved to NumPy with inline normalization for consistent dtypes.
  * Improved multi-backend shape handling for runtime compatibility.

* **Documentation**
  * Resizer docs now note broader pipeline compatibility.

* **Tests**
  * Removed a legacy normalization unit test.

[Review Change Stack](https://app.coderabbit.ai/change-stack/tue-bmd/zea/pull/352?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


wesselvannierop:
Actually, the resizing in the dataloader will always use the keras backend, changing that now.

swpenninga:
Why is `convert_to_tensor` false by default? shouldn't we be GPU-standard?

wesselvannierop:
> Why is `convert_to_tensor` false by default? shouldn't we be GPU-standard?

@tristan-deep and I thought that made it clear that the Dataloader is using cpu. But maybe you are right that True is a better default. Otherwise all our scripts will have `convert_to_tensor=True` haha

tristan-deep:
> > Why is `convert_to_tensor` false by default? shouldn't we be GPU-standard?
> 
> @tristan-deep and I thought that made it clear that the Dataloader is using cpu. But maybe you are right that True is a better default. Otherwise all our scripts will have `convert_to_tensor=True` haha

Hmm, yeah also not sure what would be best. Adding an explicit tensor conversion after your dataloader is not that bad I think. Quite transparent. 

If we choose to convert to (GPU) tensor as a last step, then also in light of #349, let's make sure we can set on which device we put the eventual tensor. Also docstring should make clear that most of dataloader will run on CPU, it is just the conversion to tensor at the end that brings it to GPU. Also what happens if you do want some GPU layers in your dataloader on one GPU, and then do model inference on another? You basically want CPU dataloader -> GPU:0 processing layer -> CPU batching etc. -> returns np.array then handoff to your script np.array -> GPU:1 model inference.

wesselvannierop:
Actually, the current  implementation of `convert_to_tensor=True` keeps the array on CPU. It just does ndarray -> whatever keras backend tensor (on cpu)

swpenninga:
isnt `convert_to_tensor` indeed the step that moves data to GPU, that only happens after? If you dont want this, you will be using back-end numpy anyways, in which case its like an identity operation?

wesselvannierop:
No clue what you mean, but the `convert_to_tensor` is executed under a `keras.device('cpu')` context manager. So the tensor will be put on cpu.

tristan-deep:
> No clue what you mean, but the `convert_to_tensor` is executed under a `keras.device('cpu')` context manager. So the tensor will be put on cpu.

In that case I would be for convert to tensor by default on actually.

swpenninga (approved):

tristan-deep (approved):

PR #353, opened 2026-05-14
Add percentile based normalization

wesselvannierop:
This is useful if your data somehow has a very large outlier that should be ignored in the normalize operation. Defaults to the old (max) behavior.


OisinNolan (dismissed):
Nice update, looks good to me!

zea/ops/tensor.py
```diff
@@ -83,12 +83,23 @@ class Normalize(Operation):
...
-    def __init__(self, output_range=None, input_range=None, **kwargs):
+    def __init__(self, output_range=None, input_range=None, percentile=None, **kwargs):
+        """
+        Args:
+            output_range (tuple): Range to which data should be mapped. Defaults to (0, 1).
+            input_range (tuple): Range of input data. If `None`, the range
+                of the input data will be computed (also see ``percentile``).
+                The tuple may contain `None` values to infer the min and / or max value from
+                the input data. Defaults to `None`.
+            percentile: Percentile to derive max value from. Defaults to None, which
```
OisinNolan: Maybe we should also mention that values above this percentile will be clipped?

wesselvannierop: I cleaned up the docstring a bit!

tristan-deep (approved):

PR #356, opened 2026-05-18
Add passband input to BandPass filter

benluijten:
Allows the user to directly set passband frequencies, independent on center_frequency.

## Summary by CodeRabbit

* **New Features**
  * Band‑pass filter can be configured with a fixed passband at setup or overridden at call time; if no fixed passband is set, demodulation frequency and bandwidth are used to derive the passband.
  * Passband selection follows clear precedence: call‑time override → fixed setup → derived values.

* **Tests**
  * Expanded tests cover init-time passband, call‑time override, protection of fixed passband, derived‑passband behavior, and invalid‑passband validation.


zea/ops/ultrasound.py
```diff
@@ -494,9 +494,9 @@ def __init__(
     def call(self, bandwidth, sampling_frequency, center_frequency, **kwargs):
         lpf = get_low_pass_iq_filter(
             self.num_taps,
-            ops.convert_to_numpy(sampling_frequency).item(),
-            ops.convert_to_numpy(center_frequency).item(),
-            ops.convert_to_numpy(bandwidth).item(),
+            sampling_frequency,
+            center_frequency,
+            bandwidth,
```
wesselvannierop: You sure it is possible to remove this (for all backends)?

benluijten: No probably not, good catch! Should be added back

tristan-deep:
@benluijten I think we can merge this after the requested changes :) Also, default behavior of the op is preserved right? Just allows for more specific passband settings.

zea/ops/ultrasound.py
benluijten: reverted in https://github.com/tue-bmd/zea/pull/356/commits/526944c9ea51b5963fbdaadbc8728f6fd373d55f

tristan-deep (approved):
Looks good!

Issue #351, opened 2026-05-11
Transition to new dataspec


tristan-deep:
Initial rewrite of dataspec is introduced in #307. 

Development of this issue is ongoing and will be merged to the branch [openh-rf](https://github.com/tue-bmd/zea/tree/openh-rf) 

## Final tasks!
- [x] Resave and reupload [carotid dataset](https://huggingface.co/datasets/zeahub/zea-carotid-2023/tree/v0.1.0) with folder structure (update script as well) @swpenninga 
- [x] Support for some custom parameters, for research purposes (#https://github.com/tue-bmd/zea/pull/307#discussion_r3258422472) @wesselvannierop #450 
- [x] Improve `zea app` robustness for dataset inspection, e.g. use lazy option for instance. @tristan-deep #451 

## Tasks after merge
- [x] After release, remove all revision=v0.1.0 tags again from zea.File in snippets / examples. @tristan-deep (have a script for this)
- [x] After do #400 @tristan-deep 

## Completed
- [x] Add legacy file support and add test for this. @swpenninga  #368 
    - [x] After all conversion scripts have been updated, also upload legacy dataset to huggingface branch v0.0.13 and new to v0.1.0. See below for assignments. @tristan-deep
- [x] Improve and clean up docs regarding data handling (zea.File, zea.Dataset, zea.Dataloader) and dataformat itself.
    - [x] General improvements @tristan-deep 
    - [x] Add doctest to getting-started.rst (see https://github.com/tue-bmd/zea/pull/307#discussion_r3123985618) #428 
- [x] Remove outdated data types figure / diagram and documentation in docs. @OisinNolan 
- [x] Check all generate_zea_dataset and remove where possible (use real legacy files for legacy compatibility instead, [see convo](https://github.com/tue-bmd/zea/pull/368#discussion_r3288172253)). @swpenninga #443 
- [x] Rename scan.py -> parameters.py @swpenninga (after #358 merge) #443 
- [x] Reduce warning stream upon `zea.File.create` (combination of ignore option and reduced / smarter warnings) @swpenninga #443 
- [x] Remove image_sc where possible (deprecated data type). @swpenninga #439 
- [x] Check the conversion examples (PICMUS etc.) and make sure we point to them in the docs / OpenH-RF page. 
    - [x] PICMUS / CETUS / CAMUS @tristan-deep #379 #385 
    - [x] DehazingEcho2025 @tristan-deep 
    - [x] EchoNetLVH @wesselvannierop #435 
[...]

Issue #328, opened 2026-04-21
Clean up data loading and saving warnings


tristan-deep:
Related to #307, #281.

Saving data erroneously will result in some errors / warnings (which is good). For instance ommitting some parameters when saving to zea dataformat results in warnings that parameter is missing and defaults are being used. Better to only throw this warning at saving time, store the default, and during loading require the parameter. Hopefully this will reduce number of printed warnings by quite a bit.

wesselvannierop:
```
zea: WARNING Optional ScanSpec field 'tgc_gain_curve' is not set. Description: Time-gain-compensation curve. Defaulted to None.
zea: WARNING Optional ScanSpec field 'waveforms_one_way' is not set. Description: One-way transmit waveforms. Defaulted to None.
zea: WARNING Optional ScanSpec field 'waveforms_two_way' is not set. Description: Two-way transmit waveforms. Defaulted to None.
zea: WARNING Optional Subject field 'age' is not set. Description:  Defaulted to None.
zea: WARNING Optional Subject field 'sex' is not set. Description:  Defaulted to None.
zea: WARNING Optional Subject field 'fat_percentage' is not set. Description:  Defaulted to None.
zea: WARNING Optional MetadataSpec field 'probe_pose' is not set. Description: Sampled probe pose at the transducer tip. Defaulted to None.
zea: WARNING Optional MetadataSpec field 'voice_narration' is not set. Description: Voice narration signal. Defaulted to None.
zea: WARNING Optional MetadataSpec field 'ecg' is not set. Description: Electrocardiogram signal. Defaulted to None.
zea: WARNING Optional MetadataSpec field 'text_report' is not set. Description: Free-text report associated with the study. Defaulted to None.
```

I feel like we shouldn't warn for every 'obscure' optional parameter that is missing.

tristan-deep:
Already discussed with @swpenninga and we are removing any warnings on loading (saving is still loud) that are about optional parameters.

wesselvannierop:
> ```
> zea: WARNING Optional ScanSpec field 'tgc_gain_curve' is not set. Description: Time-gain-compensation curve. Defaulted to None.
> zea: WARNING Optional ScanSpec field 'waveforms_one_way' is not set. Description: One-way transmit waveforms. Defaulted to None.
> zea: WARNING Optional ScanSpec field 'waveforms_two_way' is not set. Description: Two-way transmit waveforms. Defaulted to None.
[...]

PR #361, opened 2026-05-19
Add warnings for optional scan fields, fix nested warnings

swpenninga:
See #328 

Most warnings were already done during file creation and during initialization of the dataloader.

For the dataloader, missing field warnings will be raised during initialization but they do not spam the user.
The only way this is possible is if `File.scan()` is called manually in a loop, such that `ScanSpec` is constructed in a loop. 
I think this would be inefficient user-behaviour rather than something we need to account for.

---
I added warnings for all optional scan parameters [as given here](https://zea--358.org.readthedocs.build/en/358/data-acquisition.html#group-reference)
- time_to_next_transmit
- azimuth_angles
- sound_speed
- tgc_gain_curve
- element_width
- waveforms_one_way
- waveforms_two_way

and optional metadata fields
- credit
- probe_pose
- voice_narration
- ecg
- text_report
- annotations

They all default to None and this will be logged as well.

---

Additionally there are now tests for the warnings (saving, loading, subject, metadata), and i fixed some of the older tests/implementations for warnings that used a nested 
```
warnings.warn(log.warning(...))
```

---

What is still missing is the backwards compatibility of datasets warning, but this should be added with the legacy file support in #351 

## Summary by CodeRabbit

* **Bug Fixes**
  * Warnings standardized to use the app logger and improved for clarity and consistency.
  * Missing optional fields now emit descriptive warnings (including subject id) and extra custom data/metadata keys trigger warnings.
  * Loading legacy/HDF5 files now emits clearer warnings for missing or legacy-formatted scan parameters.

* **Tests**
  * Added extensive tests covering specification warnings, subject/metadata field messages, custom key warnings, and legacy/HDF5 load warnings.

[Review Change Stack](https://app.coderabbit.ai/change-stack/tue-bmd/zea/pull/361?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/tests/data/test_spec.py
+++ b/tests/data/test_spec.py
@@ -12,6 +12,7 @@ from zea.data.spec import (
     FileSpec,
     Image,
     Map,
+    MetadataSpec,
     MetricsSpec,
     ProbePose,
     ScanSpec,
@@ -444,7 +445,7 @@ def test_subject_id_warning_for_missing_id():
             },
         )
     messages = [str(c.args[0]) for c in mock_warn.call_args_list]
-    assert any("Subject ID is not provided" in m for m in messages)
+    assert any("Optional Subject field 'id' is not set" in m for m in messages)
 
 
 class TestScanValidationErrors:
@@ -837,7 +838,53 @@ class TestScanSpecSaveWarnings:
                 metadata={"subject": {"type": "human"}},
             )
         messages = [str(c.args[0]) for c in mock_warn.call_args_list]
-        assert any("Subject ID is not provided" in m for m in messages)
+        assert any("Optional Subject field 'id' is not set" in m for m in messages)
+
+
+class TestSubjectFieldWarnings:
+    """log.warning calls emitted when Subject optional fields are None."""
+
+    @pytest.mark.parametrize(
+        "field",
+        ["id", "type", "age", "sex", "fat_percentage"],
+    )
+    def test_optional_subject_field_missing_warns(self, field):
+        with patch("zea.log.warning") as mock_warn:
+            Subject()
+        messages = [str(c.args[0]) for c in mock_warn.call_args_list]
+        assert any(f"Optional Subject field '{field}' is not set" in m for m in messages)
+
+    def test_no_warning_when_all_fields_provided(self):
+        with patch("zea.log.warning") as mock_warn:
+            Subject(
+                id="patient-001",
+                type="human",
+                age=np.uint8(42),
+                sex="f",
+                fat_percentage=np.float32(17.5),
+            )
+        messages = [str(c.args[0]) for c in mock_warn.call_args_list]
+        assert not any("Optional Subject field" in m for m in messages)
+
+
+class TestMetadataSpecFieldWarnings:
+    """log.warning calls emitted when MetadataSpec optional fields are None."""
+
+    @pytest.mark.parametrize(
+        "field",
+        ["credit", "probe_pose", "voice_narration", "ecg", "text_report", "annotations"],
+    )
+    def test_optional_metadata_field_missing_warns(self, field):
+        with patch("zea.log.warning") as mock_warn:
+            MetadataSpec()
+        messages = [str(c.args[0]) for c in mock_warn.call_args_list]
+        assert any(f"Optional MetadataSpec field '{field}' is not set" in m for m in messages)
+
+    def test_no_warning_when_field_is_provided(self):
+        with patch("zea.log.warning") as mock_warn:
+            MetadataSpec(credit="Doe et al.")
+        messages = [str(c.args[0]) for c in mock_warn.call_args_list]
+        assert not any("Optional MetadataSpec field 'credit'" in m for m in messages)
 
 
 class TestLoadingWarnings:
--- a/tests/data/test_spec.py
+++ b/tests/data/test_spec.py
@@ -733,13 +733,9 @@ class TestScanSpecSaveWarnings:
     @pytest.mark.parametrize(
         "field",
         [
-            "time_to_next_transmit",
-            "azimuth_angles",
-            "sound_speed",
-            "tgc_gain_curve",
-            "element_width",
-            "waveforms_one_way",
-            "waveforms_two_way",
+            f.name
+            for f in fields(ScanSpec)
+            if f.default is None and f.name in ScanSpec.FIELD_METADATA
         ],
     )
     def test_optional_scan_field_missing_warns(self, field):
@@ -844,10 +840,7 @@ class TestScanSpecSaveWarnings:
 class TestSubjectFieldWarnings:
     """log.warning calls emitted when Subject optional fields are None."""
 
-    @pytest.mark.parametrize(
-        "field",
-        ["id", "type", "age", "sex", "fat_percentage"],
-    )
+    @pytest.mark.parametrize("field", list(Subject.SCHEMA))
     def test_optional_subject_field_missing_warns(self, field):
         with patch("zea.log.warning") as mock_warn:
             Subject()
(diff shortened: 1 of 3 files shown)


tests/data/test_spec.py
```diff
@@ -422,9 +427,7 @@ def test_schema_keys_match_dataclass_fields_for_all_specs():
 def test_subject_id_warning_for_missing_id():
     n_frames, n_tx, n_el, n_ax, n_ch = 3, 2, 4, 8, 1
 
-    with pytest.warns(
-        match="Subject ID is not provided; please consider adding an ID for better traceability"
-    ):
```
wesselvannierop, line 443: These replacements don't check the warning message anymore?

zea/data/spec.py
```diff
@@ -1177,6 +1174,24 @@ def __post_init__(self):
...
 
+        # Warn about optional fields that were not provided
+        _optional_scan_fields = {
+            "time_to_next_transmit": "time between transmit events (n_frames, n_tx)",
+            "azimuth_angles": "azimuthal angles of transmit beams (n_tx,) in radians",
+            "sound_speed": "speed of sound in m/s",
+            "tgc_gain_curve": "time-gain-compensation curve (n_ax,)",
+            "element_width": "element width of the probe in meters",
+            "waveforms_one_way": "one-way transmit waveforms (n_tx, n_samples)",
+            "waveforms_two_way": "two-way transmit waveforms (n_tx, n_samples)",
```
wesselvannierop: Can't you use `FIELD_METADATA` to avoid duplicating the description?

wesselvannierop: Additionally, maybe you can fetch the optional fields just by inspecting the dataclass fields set to None?

zea/data/spec.py
```diff
@@ -1177,6 +1174,24 @@ def __post_init__(self):
...
+        }
+        for s_field, description in _optional_scan_fields.items():
+            if getattr(self, s_field) is None:
+                log.warning(
+                    f"Optional ScanSpec field '{s_field}' is not set."
+                    f"Description: {description}. "
+                    f"Defaulted to None."
+                )
```
wesselvannierop: (see previous comment) this for loop logic can then just live on the base `Spec` class I guess.

tests/data/test_spec.py
swpenninga: it is simply replaced with patch() as mock_warn below, functionality is the same. 
this is because the pytest.warns required `warnings.warn(log.warning(...))`.

wesselvannierop: Sure, but it doesn't test what warning is emitted, just any warning will now be okay.

zea/data/spec.py
swpenninga: Good point, I didnt see it

tristan-deep:
Is the idea that we warn users mainly upon saving? And not so much upon loading?

tests/data/test_spec.py
swpenninga: https://github.com/tue-bmd/zea/pull/361/changes#diff-7c1dd3d59967eea4cd33b1dab009be99197b3bf1cff4fe9ded2cb5bbd48ff490R330

identical functionality here, right?

swpenninga:
> Is the idea that we warn users mainly upon saving? And not so much upon loading?

As described in #328 
But there is still warning on loading if something is wrong.

tests/data/test_spec.py
wesselvannierop: Sorry did not see that!

wesselvannierop:
I moved the logic to a method on the Spec class, such that it can be reused if we decide to log more of these warnings: https://github.com/tue-bmd/zea/pull/361/commits/c513e6fbfe3b5f8d375a2104c9678fd5355ec22c

wesselvannierop (dismissed):

wesselvannierop:
@coderabbitai review

swpenninga:
> I moved the logic to a method on the Spec class, such that it can be reused if we decide to log more of these warnings: [c513e6f](https://github.com/tue-bmd/zea/commit/c513e6fbfe3b5f8d375a2104c9678fd5355ec22c)

Smart, i was struggling a little since the data/scan/metadata specs varied but this seems like a general purpose solution

tristan-deep:
@coderabbitai full review

wesselvannierop (dismissed):

tests/data/test_spec.py
```diff
@@ -10,6 +12,7 @@
     FileSpec,
     Image,
     Map,
+    MetadataSpec,
```
tristan-deep, line 15: Let me test one thing before we merge

tristan-deep:
Very nice @swpenninga ! Looks clean. Something I was wondering. I'm running the example script [here](https://github.com/open-h/OpenH-RF/blob/main/examples/saving/echocardiography_example.py). Most warnings look good to me, except one. The custom field is also triggered on loading, not sure if a user could do anything about that at this point? Or is it good that people are informed it is a custom signal? In that case we want also a different warning since the suggestion doesn't make sense upon loading i think.

Specific warning I'm talking about:
```bash
zea: WARNING Custom signal key(s) added to 'metadata': probe_orientation. These are validated as generic SignalND specs. If your signal matches an existing type, prefer one of the supported signal fields: ecg, probe_pose, voice_narration.
```

All warnings: 
```bash
# saving
zea: WARNING Optional ScanSpec field 'time_to_next_transmit' is not set. Description: Time between transmit events. Defaulted to None.
zea: WARNING Optional ScanSpec field 'tgc_gain_curve' is not set. Description: Time-gain-compensation curve. Defaulted to None.
zea: WARNING Optional ScanSpec field 'element_width' is not set. Description: Element width of the probe. Defaulted to None.
zea: WARNING Optional ScanSpec field 'waveforms_one_way' is not set. Description: One-way transmit waveforms. Defaulted to None.
zea: WARNING Optional ScanSpec field 'waveforms_two_way' is not set. Description: Two-way transmit waveforms. Defaulted to None.
zea: WARNING Custom signal key(s) added to 'metadata': probe_orientation. These are validated as generic SignalND specs. If your signal matches an existing type, prefer one of the supported signal fields: ecg, probe_pose, voice_narration.
zea: WARNING Optional MetadataSpec field 'probe_pose' is not set. Description: Sampled probe pose at the transducer tip. Defaulted to None.
zea: WARNING Optional MetadataSpec field 'voice_narration' is not set. Description: Voice narration signal. Defaulted to None.
zea: File saved to echocardiography_dataset.hdf5

# loading
# Echocardiography dataset saved and reloaded successfully.
# only when calling file.metadata(), following warnings are triggered
zea: WARNING Custom signal key(s) added to 'metadata': probe_orientation. These are validated as generic SignalND specs. If your signal matches an existing type, prefer one of the supported signal fields: ecg, probe_pose, voice_narration.
zea: WARNING Optional MetadataSpec field 'probe_pose' is not set. Description: Sampled probe pose at the transducer tip. Defaulted to None.
zea: WARNING Optional MetadataSpec field 'voice_narration' is not set. Description: Voice narration signal. Defaulted to None.
```

swpenninga:
> Very nice @swpenninga ! Looks clean. Something I was wondering. I'm running the example script [here](https://github.com/open-h/OpenH-RF/blob/main/examples/saving/echocardiography_example.py). Most warnings look good to me, except one. The custom field is also triggered on loading, not sure if a user could do anything about that at this point? Or is it good that people are informed it is a custom signal? In that case we want also a different warning since the suggestion doesn't make sense upon loading i think.

You couldn't do anything about it at this point no, but note that this only happens once when you run `file.metadata()`, I think warning once is nice, since the contents deviated from expected behaviour?

tristan-deep:
> You couldn't do anything about it at this point no, but note that this only happens once when you run `file.metadata()`, I think warning once is nice, since the contents deviated from expected behaviour?

Yes I agree, but the message also says; `If your signal matches an existing type, prefer one of the supported signal fields: ecg, probe_pose, voice_narration.` That might be a bit confusing to people just reading a file? Will it be hard to change the warning from read / write? Or make the warning more explicit that this part is just about writing?

swpenninga:
Oh yeah, let me fix this

tristan-deep (approved):
🚀

tristan-deep:
I tested your latest change with custom data and metadata and worked nicely!

PR #371, opened 2026-05-21
Speed up simulation notebook pytest

### LINKED ISSUE #364: Investigate slow notebook runs
Example: https://github.com/tue-bmd/zea/actions/runs/26096594084/job/76739588740

Claude traced it back to huggingface being slow sometimes. Maybe we can mount the cache location in the local runners @benluijten ?

tristan-deep:
See #364. I saw consistency between slow pytest heavy runs where the simulation notebook stalled. Since addition of the benchmark (last cell) in this notebook I think it has been quite slow. Might not solve #364 entirely but partly maybe. There was an easy win by reducing the num of iterations in the last cell with parameter injection.

## Summary by CodeRabbit

* **Documentation**
  * Updated the simulation example notebook with improved code structure for better readability and refreshed benchmark performance data.

* **Tests**
  * Added continuous integration configuration to automatically validate the example notebook as part of the test suite.

[Review Change Stack](https://app.coderabbit.ai/change-stack/tue-bmd/zea/pull/371?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/tests/test_notebooks.py
+++ b/tests/test_notebooks.py
@@ -109,6 +109,9 @@ NOTEBOOK_PARAMETERS = {
         "grid_size_x": 16,
         "grid_size_z": 16,
     },
+    "zea_simulation_example.ipynb": {
+        "n_repeats": 2,
+    },
     # Add more notebooks and their parameters here as needed
     # "other_notebook.ipynb": {
     #     "param1": value1,
--- a/docs/source/notebooks/data/zea_simulation_example.ipynb
+++ b/docs/source/notebooks/data/zea_simulation_example.ipynb
@@ -89,15 +89,7 @@
    "execution_count": 4,
    "id": "bd7f70ad",
    "metadata": {},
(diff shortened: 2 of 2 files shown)


tests/test_notebooks.py
```diff
@@ -109,6 +109,9 @@
         "grid_size_x": 16,
         "grid_size_z": 16,
     },
+    "zea_simulation_example.ipynb": {
+        "n_repeats": 2,
```
wesselvannierop, line 113: Maybe even 1?

wesselvannierop (approved):

tests/test_notebooks.py
tristan-deep: In this case probably could indeed. I'm often hesitant to change to 1 as the behavior is often different from > 1. (Squeezing axes, initializations, etc.).

PR #372, opened 2026-05-21
Introduce a beamforming CLI

wesselvannierop:
I noticed that often I just want to quickly inspect a bmode from a `zea.File`, or even beamform an entire `zea.Dataset` to inspect which file I want to use.

I created a little CLI tool for this, which uses the speed of the new #256 `zea.Dataloader`, to load in parallel to beamforming. It also saves in parallel to beamforming. It can save to hdf5, gif or mp4.

Still needs a bit of clean-up.

### Example
```bash
python -m zea.data.process ./path/to/dataset ./output/ --config hf://zeahub/zea-cardiac-2026/config.yaml
```


tristan-deep:
Nice! That is very useful, in a way this deprecates the Interface class. Wdyt of removing that one? It is not being used or tested a lot anyways now I have the feeling. 

It also looks a lot like the backend of the openh-rf UI I made with gradio. Lets see if we can combine best of both worlds.

wesselvannierop:
> Nice! That is very useful, in a way this deprecates the Interface class. Wdyt of removing that one? It is not being used or tested a lot anyways now I have the feeling.
> 
> It also looks a lot like the backend of the openh-rf UI I made with gradio. Lets see if we can combine best of both worlds.

Indeed, I also think the interface class is not really used and I think this could be a nice alternative that suits more to the users of `zea`. When we are going to remove the interface class there is probably some more config related stuff that can be cleaned up (zea/internal/config). I would propose to do that in another PR though.

And indeed very similar to your OpenH-RF backend, this is basically a batch-optimized version of that, to do a lot in parallel to GPU processing.

tristan-deep:
Maybe we could use this branch to:
- Remove the Interface class, addressed in #412 
- Clean up file operations 
- Replace Interface __main__.py call with this beamforming CLI
- Test it all out on openh-rf-dev with the examples there to quickly process (.hdf5 + .yaml) -> (.png). Maybe even replace the gradio backend with this beamforming CLI.

Probably easiest to do this by changing target branch to openh-rf as well as it is so intertwined. WDYT @wesselvannierop ?

Other related features #374

tristan-deep:
I wanted to merge this, but was kind of hard to update it to be ready for v0.1.0. Starting again in separate PRs. There are also some unrelated changes in this PR. We can also incorporate these later. 

See:
- #415 
- #416

There might be still some nuggets of code in this PR that could be transferred.

PR #379, opened 2026-05-26
Extended ProbeSpec

### LINKED ISSUE #112: zea.data.convert.matlab.read_raw_data() requires fix for nr frames is one
In zea/data/convert/matlab.py, the function read_raw_data() does not take care well of the case where the number of frames equals 1.

Suggested addition to matlab.py (around line 520):

```python
frame_indices = get_frame_indices(file, frames)

    if raw_data.ndim == 2:
        # If there is only a single frame, we need to add a frame dimension
        raw_data = np.expand_dims(raw_data, axis=0)
    elif raw_data.ndim != 3:
        raise ValueError(
            f"Raw data has {raw_data.ndim} dimensions, but should have 2 or 3."
            "Please check the input file."
        )

    raw_data = raw_data[frame_indices]
```

### LINKED ISSUE #363: Add ProbeSpec
I'm thinking maybe we need to add a `ProbeSpec` instead of just `probe_name`, we could add:
 
- name
- bandwidth_percent*
- lens_sound_speed*
- lens_thickness*
- bandwidth_percent / bandwidth (MHz)*
- element_width*
- (move probe geometry over?)
- type (linear, phased, ...)*
- center_frequency (of the probe!)*
 
where * is currently not in the dataspec.

This would also make it easier for us to map file.scan -> Scan() and file.probe -> Probe().

### LINKED ISSUE #367: PICMUS dataset incomplete
Our conversion [script](https://github.com/tue-bmd/zea/blob/main/zea/data/convert/picmus.py) for PICMUS can be improved in the following ways:
- automatic downloading as done in CAMUS / CETUS [script](https://github.com/tue-bmd/zea/blob/main/zea/data/convert/cetus.py)
- include the in-vivo PICMUS partition which we currently do not list (and is also not on [Hugging Face Hub](https://huggingface.co/datasets/zeahub/picmus))

See download links here: https://www.creatis.insa-lyon.fr/Challenge/IEEE_IUS_2016/download

tristan-deep:
As outlined in #363, added a more detailed ProbeSpec class. Should now pass a dict of parameters (mostly optional) instead of just a probe name to `zea.File.create`. Inside a zea file you now have, `data`, `metadata`, `metrics`, `scan` and `probe` groups. The `probe` key was added in favor of the `probe_name` file attribute. `scan.probe_geometry` is now moved to `probe.probe_geometry` and will be passed to `zea.Scan` class as well when initiation a `scan=file.scan()`.

Also updated and standardized the PICMUS dataset conversion script. Does automatic downloading and uploading now and also includes the invivo partition, see #367.

Updated CAMUS and CETUS scripts as well. 

I have already uploaded a revision of the new dataset here: https://huggingface.co/datasets/zeahub/picmus/tree/v0.1.0a1

It was generated as follows:

```bash
python -m zea.data.convert picmus ./raw_picmus ./output_picmus --download --upload --revision v0.1.0a1
```

## TODO
- [x] Update conversion scripts
    - [x] PICMUS / CETUS / CAMUS (unified CLI etc.)
    - [x] EchonetLVH
    - [x] Verasonics
    - [x] EchonetDynamic
    - [x] miscellaneous (images etc.)
- [x] Update probe module? Maybe define probes not as inherited classes but as dicts now? 

## Summary by CodeRabbit

* **New Features**
  * Dataset conversion tools now support uploading to HuggingFace Hub for PICMUS, CAMUS, and CETUS datasets.

* **API Changes**
  * `file.probe`, `file.scan`, `file.metadata`, and `file.metrics` are now accessed as properties instead of method calls.
  * `File.create()` now accepts `probe={"name": "..."}` instead of `probe_name="..."`.

* **Data Format**
  * HDF5 specification reorganized with a dedicated "Probe" group, better separating probe configuration from scan parameters.


zea/data/file.py
```diff
@@ -246,15 +252,20 @@ def create(
         if path.exists() and not overwrite:
             raise FileExistsError(f"File already exists: {path}")
 
+        if probe_name is not None:
+            raise TypeError(
+                "probe_name is no longer supported. "
+                "Use probe={'name': ...} to specify the probe name."
+            )
```
wesselvannierop, line 778: Didn't we decide to not be backwards compatible with previous iterations of the spec before 0.1.0? But also fine by me if we are.

zea/data/spec.py
```diff
@@ -1165,6 +1170,159 @@ def __post_init__(self):
...
+        type: Probe geometry type: ``"linear"``, ``"phased"``, ``"curved"``, etc.
+        center_frequency: Probe nominal centre frequency in Hz.
+        bandwidth_percent: Fractional bandwidth as a percentage (0-100).
+        probe_geometry: Element positions in metres, shape (n_el, 3) with columns
+            (x, y, z).  :attr:`n_elements` and :attr:`pitch` are computed
+            automatically as read-only properties from this array.
+        element_width: Width of a single transducer element in metres.
+        element_height: Height (elevation aperture) of a single element in metres.
+        kerf: Gap between adjacent elements in metres.
+        pulse_duration: Excitation pulse duration in cycles.
```
wesselvannierop: Is this fixed for a probe? I guess this is more of a transmit thing right?

zea/data/spec.py
```diff
@@ -1165,6 +1170,159 @@ def __post_init__(self):
...
+        name: Probe model identifier (e.g. ``"verasonics_l11_4v"``).
+        type: Probe geometry type: ``"linear"``, ``"phased"``, ``"curved"``, etc.
+        center_frequency: Probe nominal centre frequency in Hz.
+        bandwidth_percent: Fractional bandwidth as a percentage (0-100).
+        probe_geometry: Element positions in metres, shape (n_el, 3) with columns
+            (x, y, z).  :attr:`n_elements` and :attr:`pitch` are computed
+            automatically as read-only properties from this array.
+        element_width: Width of a single transducer element in metres.
+        element_height: Height (elevation aperture) of a single element in metres.
+        kerf: Gap between adjacent elements in metres.
```
wesselvannierop: We can compute this from geometry and element width no?

zea/data/spec.py
```diff
@@ -1165,6 +1170,159 @@ def __post_init__(self):
...
+    @property
+    def pitch(self) -> float | None:
+        """Centre-to-centre element spacing in metres, derived from :attr:`probe_geometry`.
+
+        Returns ``None`` when *probe_geometry* is not set or has only one element.
+        """
+        if self.probe_geometry is not None and self.probe_geometry.shape[0] > 1:
+            x_span = float(self.probe_geometry[-1, 0] - self.probe_geometry[0, 0])
+            return x_span / (self.probe_geometry.shape[0] - 1)
+        return None
```
wesselvannierop, line 1387: Assumes 2D ULA

tristan-deep: good point, remove method, or add like if probe type is linear compute this else notimplementederror?

zea/data/spec.py
tristan-deep: indeed it is redundant, i'll remove it

zea/data/spec.py
tristan-deep: correct, i'll remove it

zea/data/file.py
tristan-deep: Yeah I agree. I wasn't sure myself if probe_name was also an attribute in legacy files? In any case this raises an error so in a way it is not backwards compatible.

zea/data/spec.py
wesselvannierop: Maybe also needs to check if the element spacing is uniform?

tristan-deep: Is checked now, maybe we should check anyways if probe type (linear) satisfies some of the other settings as not to be conflicting.

tristan-deep:
@OisinNolan could you resolve merge conflicts here? I think mostly track related conflicts. I already resolved main -> openh-rf [just now](https://github.com/tue-bmd/zea/pull/382), so once merge conflicts are resolved here I think we are up to date.

zea/data/file.py
```diff
@@ -199,7 +204,9 @@ def create(
                 :class:`~zea.data.spec.MetadataSpec`.
             metrics: Optional metrics dict accepted by
                 :class:`~zea.data.spec.MetricsSpec`.
-            probe_name: Name of the probe.
+            probe_name: Removed — use ``probe={'name': ...}`` instead.
```
OisinNolan, line 715: Do we still want to be able to create a new file with a probe_name? Or rather just support loading files with a probe_name? (I see this is already handled by FileSpec.from_hdf5)

OisinNolan: Ah I see that an error is thrown if probe_name is not None -- I guess that will make it easier to update the example scripts for the new pre-release. Once they're updated we could then consider removing probe_name from `.create`

wesselvannierop: Indeed

OisinNolan (commented):
I wonder if it could be confusing to users that we have in some cases the same attributes on both scan and probe? e.g. [bandwidth_percent](https://github.com/tue-bmd/zea/blob/7599d3ff3e53bbb2922b9239850160aed8632140/zea/scan.py#L214) -- should the scan bandwidth_percent be interpreted as the bandwidth used for this particular sequence, or should it match the probe's bandwidth percent (= the maximum possible bandwidth of the probe?). There are also some discrepancies, e.g. on ProbeSpec the docstring [says the max value should be 100](https://github.com/tue-bmd/zea/blob/7599d3ff3e53bbb2922b9239850160aed8632140/zea/data/spec.py#L1273), but if you don't specify any value for probe.bandwidth_percent, then scan.bandwidth_percent will default to 200.

I would generally be in favor of having scan-varying params in the scan, keeping the probe params as fixed, and updating naming to make that clear so that there isn't any overlap between the scan param names and probe param names. What do you guys think? We could still use scan params to populate probe for legacy file support.

tristan-deep:
> I wonder if it could be confusing to users that we have in some cases the same attributes on both scan and probe? e.g. [bandwidth_percent](https://github.com/tue-bmd/zea/blob/7599d3ff3e53bbb2922b9239850160aed8632140/zea/scan.py#L214) -- should the scan bandwidth_percent be interpreted as the bandwidth used for this particular sequence, or should it match the probe's bandwidth percent (= the maximum possible bandwidth of the probe?). There are also some discrepancies, e.g. on ProbeSpec the docstring [says the max value should be 100](https://github.com/tue-bmd/zea/blob/7599d3ff3e53bbb2922b9239850160aed8632140/zea/data/spec.py#L1273), but if you don't specify any value for probe.bandwidth_percent, then scan.bandwidth_percent will default to 200.
> 
> I would generally be in favor of having scan-varying params in the scan, keeping the probe params as fixed, and updating naming to make that clear so that there isn't any overlap between the scan param names and probe param names. What do you guys think? We could still use scan params to populate probe for legacy file support.

No I agree with you actually, probe parameters should be fixed properties in a way. I don't think we need legacy support for this. The scan group in the file can store the acquisition varying properties. The scan class is more complicated however, it might require some of the probe parameters for its internal computations.

We should also resolve the discrepancies you mention in this PR.

wesselvannierop:
Good catch on the docstring of bandwidth_percent, this can for sure be >100.

I think, in a future PR, we should make the Probe class an attribute on the Scan class. We can for legacy reasons add the Probe attributes as properties on the Scan class such as:

```python
@property
def bandwidth_percent(self):
    """Fractional bandwidth as a percentage."""
    return self.probe.bandwidth_percent
```

tristan-deep:
> Good catch on the docstring of bandwidth_percent, this can for sure be >100.
> 
> I think, in a future PR, we should make the Probe class an attribute on the Scan class. We can for legacy reasons add the Probe attributes as properties on the Scan class such as:
> 
> ```python
> @property
> def bandwidth_percent(self):
>     """Fractional bandwidth as a percentage."""
>     return self.probe.badwidth_percent
> ```

I was also thinking about such a solution, and also immediately good suggestion for legacy support!

tristan-deep:
I think all conversion scripts have been updated as such that they at least run. We can make another PR to actually improve / update them further as to keep this PR contained to mainly the ProbeSpec update.

tristan-deep:
I implemented some legacy fixes (introduces by the updated ProbeSpec), building on this PR in #385.

zea/data/file.py
```diff
@@ -661,7 +712,9 @@ def create(
...
+            probe: Probe specification as a :class:`~zea.probes.Probe` object or a
+                plain dict accepted by :class:`~zea.data.spec.ProbeSpec`.
```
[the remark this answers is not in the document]
tristan-deep: zea.Probe is a subclass of zea.data.spec.ProbeSpec, won't that just pass?

zea/data/convert/cetus.py
```diff
@@ -211,7 +231,11 @@ def process_cetus(source_path, output_path, overwrite=False):
             "credit": CETUS_CITATION,
             "annotations": {"label": np.array([time_point])},
         },
-        probe_name="generic",
+        probe={"name": "generic"},
+        # TODO: includes files from both:
+        # - GE system refered as Vivid E9, using a 4V probe;
+        # - Philips system refered as iE33, using a X5-1 probe;
+        # - Siemens system refered as SC2000, using 4Z1c probe.
```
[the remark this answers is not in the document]
tristan-deep: Should be resolved now.

zea/data/convert/utils.py
```diff
@@ -136,6 +136,48 @@ def unzip(src: str | Path, dataset: str) -> Path:
...
+        destination: Full file path where the downloaded content will be saved.
+            The parent directory is created if it does not exist.
+
+    Returns:
+        Path to the (possibly pre-existing) downloaded file.
+    """
+    destination = Path(destination)
+    if destination.exists():
+        log.info("File already exists: %s. Skipping download.", destination.name)
+        return destination
```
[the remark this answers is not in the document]
tristan-deep: Should be resolved now.

zea/data/convert/verasonics.py
```diff
@@ -1220,6 +1146,132 @@ def to_zea(
...
+            element_width = element_width * self.wavelength
+
+        return element_width
+
+    @property
+    def element_length(self):
+        """Element length for row-column probes in meters."""
+        if "ElementLength" in self.trans_obj.keys():
+            return self.trans_obj["ElementLength"][:].item() * 1e6
+
```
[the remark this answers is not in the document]
tristan-deep: @wesselvannierop ?

zea/data/file_operations.py
```diff
@@ -146,8 +146,8 @@ def save_file(
         data=data,
         scan=scan_dict if scan_dict else None,
         metadata=metadata or None,
-        probe_name="generic",
-        description=description,
+        probe={"name": "generic"},
+        description=description or None,
```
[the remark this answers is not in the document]
tristan-deep: Should be resolved now.

tests/test_probes.py
```diff
@@ -34,58 +35,52 @@ def test_get_default_scan_paramters(probe_name):
...
+        "demodulation_frequency": np.float32(6.25e6),
+        "initial_times": np.zeros(n_tx, dtype=np.float32),
+        "t0_delays": np.zeros((n_tx, n_el), dtype=np.float32),
+        "tx_apodizations": np.ones((n_tx, n_el), dtype=np.float32),
+        "focus_distances": np.full(n_tx, np.inf, dtype=np.float32),
+        "transmit_origins": np.zeros((n_tx, 3), dtype=np.float32),
+        "polar_angles": np.zeros(n_tx, dtype=np.float32),
+        "time_to_next_transmit": np.ones((n_frames, n_tx), dtype=np.float32) * 1e-4,
+    }
+    _, path = tempfile.mkstemp(suffix=".hdf5")
```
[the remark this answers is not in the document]
tristan-deep: Should be resolved now.

docs/source/spec_doc.py
```diff
@@ -404,7 +405,16 @@ def generate() -> str:
...
+    # ---- probe tab -----------------------------------------------------------
+    lines += [
+        "   .. tab-item:: probe",
+        "      :sync: probe",
+        "",
+        "      Probe group with probe geometry and frequency parameters.",
+        "",
+        rst_full_table(ProbeSpec, base_indent=2),
+        "",
+    ]
```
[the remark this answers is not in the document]
tristan-deep: Should be resolved now.

docs/source/_spec_ref.rst
```diff
@@ -765,6 +753,76 @@ Fields marked :bdg-secondary:`optional` may be absent; all others are
...
+           - m/s
+           - Speed of sound in the acoustic lens.
+           - |badge-opt|
+         * - ``lens_thickness``
+           - ``float32``
+           - scalar
+           - m
+           - Thickness of the acoustic lens.
+           - |badge-opt|
+
```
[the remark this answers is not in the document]
tristan-deep: Should be resolved now.

docs/source/notebooks/data/zea_data_example.ipynb
```diff
@@ -550,7 +550,8 @@
    "source": [
     "with zea.File(config.data.dataset_folder + \"/\" + config.data.file_path, mode=\"r\") as file:\n",
     "    # we use config here to overwrite some of the scan parameters\n",
-    "    scan = file.scan(**config.scan)\n",
+    "    scan = file.scan\n",
+    "    scan.update(**config.scan)\n",
```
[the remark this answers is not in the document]
wesselvannierop: Will fix this later.

wesselvannierop (approved):

PR #383, opened 2026-05-27
Make aligned data spatial map & Allow 2D segmentations

swpenninga:
See tickboxes of #351 

- [x] Added aligned data spatial map (made similar to beamforming map)
- [x] Enable 2D segmentations

### DIFF
--- a/zea/data/spec.py
+++ b/zea/data/spec.py
@@ -686,6 +686,44 @@ class Image(Map):
 
 
 @dataclass
+class AlignedData(Spec):
+    """Time-of-flight corrected data.
+
+    Args:
+        values: The aligned data of shape ``(n_frames, n_tx, n_ax, n_el, n_ch)``
+            and type float32 or int16. n_ch is 1 for RF data or 2 for IQ data.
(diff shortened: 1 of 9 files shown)


tristan-deep (approved):
🚀 nice worked for me (tested 2D segmentation maps and saving aligned data)

zea/data/spec.py
```diff
@@ -659,18 +660,34 @@ class Segmentation(BooleanMap):
...
+    SCHEMA = {
+        **BooleanMap.SCHEMA,
+        "values": {
+            **BooleanMap.SCHEMA["values"],
+            "shape": (
+                ("n_frames", "z", "x", "y", "n_spatial_ch"),
+                ("n_frames", "z", "x", "n_spatial_ch"),
+            ),
+        },
+    }
```
wesselvannierop, line 682: Maybe this should be a shape that `Map` accepts?

wesselvannierop: Actually no, this is better!

wesselvannierop (approved):

PR #393, opened 2026-05-29
Split heavy tests

swpenninga:
Regarding #364 

Does not address the problem mentioned, but lowers the cost when tests fail. If a single test fails, we dont need to re-run the whole 30min+, this lowers the burden/bottlenecking on the runners.

Current proposed split:

- `heavy-tests` contains `ops_infra, transmit_schemes, conversion_scripts`
- `notebook-tests-agent` 2 notebooks
- `notebook-tests-models` 9 model notebooks
- `notebook-tests-pipeline` 7 notebooks
- `notebook-tests-data-metrics` 5 very cheap notebooks

## Summary by CodeRabbit

* **Tests**
  * Reorganized heavy test execution to target specific test sets for more focused runs
  * Split notebook tests into separate parallel jobs (agent, models, pipeline, data/metrics) to speed CI
  * Added per-notebook timing collection and grouped folder summaries to improve visibility into test performance and totals
  * Added option to run notebook tests filtered by directory for targeted runs
  * Improved test cache isolation by using per-test temporary cache handling

* **Chores**
  * CI updated to mount shared cache volumes, set in-container cache env vars, and renamed the heavy-tests step for clarity

### DIFF
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -137,7 +137,7 @@ jobs:
             -v "${{ github.workspace }}:/zea" \
             -w /zea \
             ${{ needs.image.outputs.docker_image }} \
-            sh -lc "pytest --cov --cov-branch --cov-report=xml -p no:cacheprovider -m 'heavy' ./tests/test_ops_infra.py ./tests/test_transmit_schemes.py ./tests/data/test_conversion_scripts.py"
+            sh -lc "pytest --cov --cov-branch --cov-report=xml -p no:cacheprovider -m 'heavy' ./tests"
 
       - name: Upload heavy test results to Codecov
         uses: codecov/codecov-action@v6
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -10,8 +10,10 @@ env:
   REGISTRY: ghcr.io
   BASE_IMAGE: ghcr.io/tue-bmd/zea/all-cpu
   MAIN_REPO: tue-bmd/zea
-  HF_CACHE_MOUNT: /opt/hf-cache:/root/.cache/huggingface
-  ZEA_CACHE_MOUNT: /opt/zea-cache:/root/.cache/zea
+  HF_CACHE_HOST: /cache/huggingface
+  HF_CACHE_CONTAINER: /cache/huggingface
+  ZEA_CACHE_HOST: /cache/zea
+  ZEA_CACHE_CONTAINER: /cache/zea
 
 jobs:
   tag:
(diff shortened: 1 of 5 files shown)


tristan-deep:
A better notebook summary (run times) would also be appreciated. Could you include that in this PR?

tristan-deep:
It does now need to spin up a new image for every group? Also if the notebook tests fail we need to manage all of them. Maybe better to add them as separate steps?

swpenninga:
Note on the new summary: 
- Github actions now has 4 notebook sessions, so the results and timings will be split into 4.
- If you just run pytest -m notebook, you will get a single summary with the 4 folders combined

swpenninga:
> It does now need to spin up a new image for every group? Also if the notebook tests fail we need to manage all of them. Maybe better to add them as separate steps?

I get that it looks like that currently, but the image job runs once, and all 5 heavy test jobs just pull the image. 

There's this weird thing going on where the single `heavy-tests` icon you see turns into 5 separate jobs once one of them is picked up, then they separate and you can rerun them individually if they fail.

swpenninga:
image

tristan-deep (approved):
I do like the splitting of the notebooks and especially separating them from the other heavy tests. It should be noted that the time from start of the test (pulling the image) to the first notebook cell starting takes approximately 2min. So this will add a 2min running time for each notebook. There is probably a trade-off between being able to divide the number of jobs and assign to different runners (especially when we will add more local-hosts or runners), and the added overhead of pulling the image for each of these jobs. For now I think this is a great solution and we can reconsider in the future to reduce the number of notebook splits again. 

You added splits of course mainly for lower costs if one fails. For instance in the case where there are bugs in the commits that break the tests. As outlined in #364 notebooks mainly failed because of a timeout. I also tried to address this by adding persistent HF / zea cache dirs across runners that we locally host. These were mainly changes to the local host server management repo, but also a few lines here:
- Had to change a few lines in the workflow file (mount volumes to the docker commands for local-host)
- Remove the temp global cache dir change, just have a temp cache dir for the cache related tests

When I run the following on local-host, I see the mounted cache dirs are being populated:
```bash
watch -n 2 "echo '--- PVC DISK USAGE ---' && du -sh /opt/hf-cache /opt/zea-cache && echo '--- FILE COUNT ---' && find /opt/hf-cache -type f | wc -l"

--- DISK USAGE ---
8.7G    /opt/hf-cache
2.9G    /opt/zea-cache
--- FILE COUNT ---
179
```

Seems to work for now, and actually fixes #364. Time will tell if that is really the case.

PR #395, opened 2026-05-29
🚨 Less warnings file and scan loading

tristan-deep:
See #328 

### What changed
- Optional `Spec` fields no longer warn at construction/load time.
- Optional field warnings are emitted when serializing (saving) only.
- This reduces noise when reading files while still surfacing incomplete metadata when writing datasets.

### Why
- Loading older or partial datasets often leaves optional fields unset by design.
- Repeated warnings during read workflows are noisy and not actionable.
- Save-time warnings are the right moment to signal missing optional metadata.

You'll also see I reran one notebook and the warnings got reduced significantly.

<details closed>
<summary>Minimal snippet</summary>

from pathlib import Path
import numpy as np
import zea

def make_minimal_scan(n_frames=2, n_tx=2, n_el=4):
    return {
        "sampling_frequency": np.float32(30e6),
        "center_frequency": np.float32(5e6),
        "demodulation_frequency": np.float32(5e6),
        "initial_times": np.zeros((n_tx,), dtype=np.float32),
        "t0_delays": np.zeros((n_tx, n_el), dtype=np.float32),
        "tx_apodizations": np.ones((n_tx, n_el), dtype=np.float32),
        "focus_distances": np.zeros((n_tx,), dtype=np.float32),
        "transmit_origins": np.zeros((n_tx, 3), dtype=np.float32),
        "polar_angles": np.zeros((n_tx,), dtype=np.float32),
        "azimuth_angles": np.zeros((n_tx,), dtype=np.float32),
        # Intentionally omit optional fields like tgc_gain_curve, waveforms_*
    }

path = Path("temp/spec_warning_save_load_demo.hdf5")
data = {"raw_data": np.zeros((2, 2, 8, 4, 1), dtype=np.float32)}
scan = make_minimal_scan()
metadata = {}

print("\n--- SAVE PHASE (expect optional warnings) ---")
zea.File.create(path, data=data, scan=scan, metadata=metadata, overwrite=True)

print("\n--- LOAD PHASE (expect no optional warnings) ---")
with zea.File(path) as f:
    _ = f.data.raw_data.shape
    _ = f.metadata
    scan = f.scan
</details>

## Summary by CodeRabbit

* **Bug Fixes**
  * Warnings for missing transmit parameters now emit only once per scan instance, reducing redundant notifications during multiple accesses.
  * Warnings about unset optional fields now appear when saving data to disk rather than during object initialization.

### DIFF
--- a/docs/source/notebooks/pipeline/zea_sequence_example.ipynb
+++ b/docs/source/notebooks/pipeline/zea_sequence_example.ipynb
@@ -168,7 +168,6 @@
     "\n",
     "scan.set_transmits(n_tx)  # reduce number of transmits for faster processing\n",
     "scan.zlims = (0, 0.04)  # reduce z-limits a bit for better visualizations\n",
-    "scan.n_ch = data.shape[-1]  # rf data\n",
     "\n",
     "config = zea.Config(dynamic_range=(-40, 0))"
    ]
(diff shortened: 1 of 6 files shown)


swpenninga (approved):
Yeah that solves it 👍

PR #402, opened 2026-06-02
Rework jax rng handling to fix failing tests

### LINKED ISSUE #401: Tests fail on jax due to stateful random seeds
The following tests currently fail on jax, all due to random seed-related issues:

```
tests/test_augmentations.py
tests/test_flow_matching.py
tests/test_generative.py
```

This was originally fixed in #373, but that made the PR too big to review, so I'm splitting off those changes into a separate PR.

Louisvh:
Previously, there were a number of places where the jax back-end would pass a stateful SeedGenerator instead of a stateless seed. This should fix that problem, but needs careful testing. If the behavior remains correct, this closes #401. 

## Summary by CodeRabbit

* **New Features**
  * Improved random seed handling for tensor operations, including support for integer seeds and broader JAX key detection.
  * Randomized augmentation now handles per-item randomness more consistently when processing batched symbolic inputs.

* **Bug Fixes**
  * Fixed seed splitting so repeated draws are reproducible while remaining independent across splits.
  * `None` seeds continue to produce valid random outputs with the expected shapes across backends.

* **Documentation**
  * Updated internal test coverage and descriptions for seed-splitting behavior.


tristan-deep:
@Louisvh could you look at this comment? I will merge this PR in the meantime into #373 so I can test the entire setup.

Louisvh:
Yeah the comment looks correct, according to the documentation iota only supports an integer-esque size (although apparently it doesn't consistently crash and fail if it _does_ get a symbolic input). The test that tests the symbolic path does so with a tensorflow symbolic, so the jax path might actually be failing. The coderabbit suggestion looks good, though we might want to change the axis name to something that's less likely to clash with a user-defined axis ("circle_batch"?).

PR #406, opened 2026-06-03
Add preset for 12ch flowmatching model

swpenninga:
## Summary by CodeRabbit

* **New Features**
  * Added a new flow matching preset for EchoNet LVH with 12-channel support, providing an additional model configuration option for flow matching applications.

### DIFF
--- a/zea/models/presets.py
+++ b/zea/models/presets.py
@@ -183,6 +183,18 @@ flow_matching_presets = {
         },
         "hf_handle": "hf://zeahub/flowmatching-echonetlvh/3ch",
     },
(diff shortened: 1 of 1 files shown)


tristan-deep (approved):
Sweet! Maybe a model card on zeahub would be nice at this point also to explain the different models in the folder. We can think of a good model card template for all our models as well.

swpenninga:
> Sweet! Maybe a model card on zeahub would be nice at this point also to explain the different models in the folder. We can think of a good model card template for all our models as well.

https://huggingface.co/zeahub/flowmatching-echonetlvh/blob/main/README.md

PR #408, opened 2026-06-03
Redesign config schema

swpenninga:
Addresses #407

## Redesign Config validation as dataclass-based Specs

Replaces the `schema`-library validation layer with dataclass-based Specs, aligning Config validation with the `ProbeSpec`/`ScanSpec` pattern used elsewhere in the codebase. The `parameters` section is now open/extensible **by design** rather than via the interim `ignore_extra_keys=True` workaround.

### What changed

**`zea/internal/config/validation.py`** — full rewrite:
- New `ConfigSpec` base class (dataclass + `__post_init__` validation + `from_dict`/`to_dict` + introspection helpers `field_names`/`required_fields`/`all_field_paths`), mirroring the `Spec` style but for config values rather than numpy arrays.
- Section specs `DataConfig`, `PlotConfig`, `PipelineConfig`, `ParametersConfig`, and top-level `ConfigSchema`, with every field/default/validator ported 1:1.
- Small validator helpers (`enum`, `regex`, `any_of`, `optional`, numeric/range checks) replacing `Or`/`And`/`Regex`.
- The `parameters` section and the top-level config are **open by design** (`ALLOW_EXTRA = True`) — arbitrary keys are stored and re-emitted unchanged, mirroring `zea.Parameters._custom_params`. This replaces the `ignore_extra_keys=True` workaround, which is gone.
- Deprecated `postprocess_schema` removed (no consumers).
- New `validate_config(dict) -> dict` entry point.

**`zea/config.py`** — `check_config` now calls `validate_config` instead of `config_schema.validate`.

**`zea/internal/config/create.py`** — interactive creator rewritten to traverse the specs via introspection; input coerced with `yaml.safe_load` + validators. Also fixed its already-broken `check_config` import (now from `zea.config`).

**`docs/source/parameters_doc.py`** — `flatten_schema_keys` replaced by `ConfigSchema.all_field_paths()`; deprecated `scan` alias tolerated via a `DEPRECATED_KEYS` set.

**Tests** — new `tests/test_config_validation.py` (defaults, required-field errors, enum/range/regex rejection, valid devices, parameters & top-level pass-through, idempotency, `scan→parameters` migration, freeze behavior); `tests/test_configs.py` updated to drop the `schema` import.

### Note on the `schema` dependency
`schema >=0.7` is left in `pyproject.toml`: it's still used by `zea/data/convert/verasonics.py` (unrelated to config), so removing it isn't safe. Nothing under `zea/`, `docs/`, or `tests/` imports `schema` for config purposes anymore.


tristan-deep:
@swpenninga could you rebase this? After I'll have a look at the changes.

swpenninga:
> @swpenninga could you rebase this? After I'll have a look at the changes.

See #412

PR #417, opened 2026-06-09
Tissue suppression

vincentvdschaft:
This PR implements singular value decomposition based tissue suppression. It implements the function `tissue_suppression`, which is used in a new operation: `TissueSuppression`.

# Result
## Without tissue suppression

verasonics_bmode

## With tissue suppression
verasonics_bmode copy

# How to test
- I placed the source data for this example (from the PALA dataset) on the NAS `USBMD-NAS/Ultrasound-BMd/data/vincent/RF_003.hdf5`
- This is the pipeline I used
```yaml
parameters:
  grid_size_x: 400
  grid_size_z: 600
  dynamic_range: [-60, 0]
  zlims: [0.001, 0.013]
  apply_lens_correction: false
  lens_sound_speed: 1000
  lens_thickness: 0.002

pipeline:
  operations:
    - name: keras.ops.cast
      params:
        dtype: float32
    - name: beamform
      params:
        beamformer: delay_and_sum
        num_patches: 200
    - name: tissue_suppression
    - name: envelope_detect
    - name: normalize
    - name: log_compress
```

## Summary by CodeRabbit

* **New Features**
  * Added tissue suppression functionality available as both a functional API (`suppress_tissue`) and an operation (`TissueSuppression`) for ultrasound signal processing.
  * Both implementations include a configurable cutoff parameter (default: 5).

* **Tests**
  * Added comprehensive test coverage for tissue suppression, verifying output shape/dtype preservation and signal reduction.

### DIFF
--- a/zea/func/ultrasound.py
+++ b/zea/func/ultrasound.py
@@ -930,7 +930,7 @@ def dehaze_nuclear_diffusion(
     return tissue_frames, haze_frames
 
 
-def suppress_tissue(data, cutoff: int = 5) -> np.ndarray:
+def suppress_tissue(data, cutoff: int = 5):
     """
     Suppresses tissue using Direct SVD.
 
@@ -940,6 +940,8 @@ def suppress_tissue(data, cutoff: int = 5) -> np.ndarray:
     """
     if cutoff <= 0:
         return data
+    if cutoff >= data.shape[0]:
+        raise ValueError(f"Cutoff must be between 0 and n_frames-1, got {cutoff}.")
 
     original_shape = data.shape
     n_frames = original_shape[0]
--- a/zea/func/ultrasound.py
+++ b/zea/func/ultrasound.py
@@ -935,7 +935,7 @@ def suppress_tissue(data, cutoff: int = 5) -> np.ndarray:
     Suppresses tissue using Direct SVD.
 
     Args:
-        data (np.ndarray): Shape (Frames, ...)
+        data (ops.Tensor): Shape (n_frames, ...)
         cutoff (int): Number of principal components (tissue) to reject.
     """
     if cutoff <= 0:
--- a/zea/func/ultrasound.py
+++ b/zea/func/ultrasound.py
@@ -946,9 +946,9 @@ def suppress_tissue(data: np.ndarray, cutoff: int = 5) -> np.ndarray:
 
     original_shape = data.shape
     n_frames = original_shape[0]
-    data_2d = data.reshape(n_frames, -1)
+    data_2d = ops.reshape(data, (n_frames, -1))
 
-    casorati_matrix = data_2d @ data_2d.T
+    casorati_matrix = ops.matmul(data_2d, ops.transpose(data_2d))
 
     # We call the data X
     # X = U @ S @ Vh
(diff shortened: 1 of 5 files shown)


zea/func/ultrasound.py
```diff
@@ -928,3 +928,39 @@ def _nuclear_diffusion_posterior_sample(
...
 
     return tissue_frames, haze_frames
+
+
+def suppress_tissue(data: np.ndarray, cutoff: int = 5) -> np.ndarray:
+    """
+    Suppresses tissue using Direct SVD.
+
+    Parameters
+    ----------
```
tristan-deep: make sure to have google style docstrings (not numpy) for the added funcs

zea/func/ultrasound.py
```diff
@@ -928,3 +928,39 @@ def _nuclear_diffusion_posterior_sample(
         haze_frames = hazy_np - haze_frames - 1
 
     return tissue_frames, haze_frames
+
+
+def suppress_tissue(data: np.ndarray, cutoff: int = 5) -> np.ndarray:
```
tristan-deep: can remove the type annotation for data for now. it is not np.ndarray but more like tensor i would say. But we have to figure out that too for the other zea.func.ultrasound funcs. Something for another PR.

zea/ops/ultrasound.py
```diff
@@ -1184,3 +1185,35 @@ def call(
...
+    def __init__(self, cutoff: int = 5, **kwargs):
+        super().__init__(**kwargs)
+        self.cutoff = cutoff
+
+    @staticmethod
+    def suppress_tissue(data: np.ndarray, cutoff: int = 5) -> np.ndarray:
+        """
+        Suppresses tissue using Direct SVD (Reduced).
+
+        Parameters
```
tristan-deep: change to google style docstring again

zea/ops/ultrasound.py
```diff
@@ -1184,3 +1185,35 @@ def call(
...
+    Removes stationary tissue components from multi-frame ultrasound data
+    by zeroing the dominant singular values of the Casorati matrix.
+    """
+
+    def __init__(self, cutoff: int = 5, **kwargs):
+        super().__init__(**kwargs)
+        self.cutoff = cutoff
+
+    @staticmethod
+    def suppress_tissue(data: np.ndarray, cutoff: int = 5) -> np.ndarray:
```
tristan-deep: again remove type annotation for data

zea/ops/ultrasound.py
```diff
@@ -1184,3 +1185,35 @@ def call(
...
+        """
+        Suppresses tissue using Direct SVD (Reduced).
+
+        Parameters
+        ----------
+        data : np.ndarray
+            Shape (Frames, Transmits, Samples, Channels)
```
tristan-deep: can use our convention for shape (n_frames, n_tx, n_ax, n_ch)

zea/ops/ultrasound.py
```diff
@@ -1184,3 +1185,35 @@ def call(
...
+class TissueSuppression(Operation):
+    """Tissue suppression using SVD-based clutter filtering.
+
+    Removes stationary tissue components from multi-frame ultrasound data
+    by zeroing the dominant singular values of the Casorati matrix.
+    """
+
+    def __init__(self, cutoff: int = 5, **kwargs):
+        super().__init__(**kwargs)
+        self.cutoff = cutoff
```
tristan-deep, line 1200: is cutoff a static or dynamic parameter (i.e. do we pass it to init of the op or during the call?). now seems to be both but the static is never used.

tristan-deep (changes requested):
Really useful!

zea/func/ultrasound.py
vincentvdschaft: Good point. I have changed it.

zea/func/ultrasound.py
vincentvdschaft: Good catch. Fixed.

zea/ops/ultrasound.py
vincentvdschaft: Good catch. It should be static I think because it determines the shape of some intermediate tensors in the suppress_tissue function.

zea/ops/ultrasound.py
vincentvdschaft: Check

tristan-deep:
@coderabbitai full review

tristan-deep (approved):
🌽

PR #418, opened 2026-06-10
Fix keras import for `--help`

swpenninga:
Regarding #416 

importing `ProcessArgs` from `zea.data.process` was causing the cascading keras import when using `zea --help`.

This is now moved to its own file. Its kind of an ugly fix, but its also the smallest possible fix.
Maybe we can think of something else, but for now, `zea --help` is clean and runs in ~1ms instead of ~1s


tristan-deep:
Would it make sense to add the other arg parse utils / funcs there as well? From the zea app which are now in main? Just so everything is together. Otherwise perfect solution!

tristan-deep (approved):
Nice!

PR #420, opened 2026-06-10
Hot fixing main

tristan-deep:
`preset_utils.py`: `_get_snapshot_dir_from_downloaded_file` now walks up the downloaded file's own path to find its snapshot directory — instead of scanning all snapshots and picking the most recently modified one. That max(mtime) approach caused runners to grab the wrong snapshot version (v0.1.0 snapshot left by openh-rf tests had a fresher mtime than the main snapshot). Basically different runners were fighting over the same cache on local host. Added revision options as well, which are probably not necessary for now, they will be anyways introduced in #358.

## Summary by CodeRabbit

* **New Features**
  * Added optional `revision` parameter to specify Hugging Face dataset versions when loading datasets.

* **Bug Fixes**
  * Forwarded revision/cache parameters for Hugging Face path resolution.
  * Improved error reporting for missing dataset keys.
  * Fixed snapshot detection logic for downloaded Hugging Face files to pick the correct snapshot.

* **Documentation**
  * Clarified `clear_cache` docstring about what is and isn’t removed.

* **Chores**
  * Updated ignore patterns to exclude additional generated/tool artifacts.


swpenninga (approved):
fix hot fix fixed hot fix

PR #426, opened 2026-06-11
Allow omitting last `time_to_next_transmit`

Felixduelmer:
Timestamps are now created by walking the global track_schedule event by event, writing the current cumulative time for each transmit and only advancing when another transmit follows. This means `time_to_next_transmit` no longer needs to include a final unused interval, and it can be provided either as the existing (n_frames, n_tx) array or as a flat interval array with n_events or n_events - 1 values.


OisinNolan (changes requested):
Thanks for this PR @Felixduelmer, nice work! Could you pls add a PR description to help review, in particular covering the changes made to `_compute_all_track_timestamps`? Thanks!

Felixduelmer:
@OisinNolan, thanks. Sorry, updated the description.

zea/data/file.py
```diff
@@ -368,7 +368,7 @@ def _compute_all_track_timestamps(
 
     Walks the schedule once, keeping a single scalar cumulative timestamp and
     a per-track ``[frame_idx, tx_idx]`` counter to index into each track's
-    ``time_to_next_transmit`` array.  The schedule must cover exactly
```
OisinNolan, line 371: any particular reason to change the naming here?

zea/data/spec.py
```diff
@@ -1141,8 +1141,9 @@ class ScanSpec(Spec):
             shape (n_tx, 3). This is the (x, y, z) position from which the beam
             is transmitted.
         polar_angles: The polar angles in radians of the transmit beams of shape (n_tx,).
-        time_to_next_transmit: The time in s between subsequent transmit events
-            of shape (n_frames, n_tx).
+        time_to_next_transmit: The time in s between subsequent transmit events.
+            Shape is either (n_frames, n_tx), or flat (n_timing_intervals,)
+            where n_timing_intervals is n_frames * n_tx or one fewer.
```
OisinNolan: I think it would be nice to have a known shape for this tensor in the file, rather than allowing the shape to be ambiguous -- curious what your thoughts are.

What do you think about allowing the user to pass `time_to_next_transmit` as an N-1 length flattened list, but then internally padding it with a `0.0` and reshaping to `n_frames, n_tx`, such that the tensor is always saved the same way?

zea/data/file.py
Felixduelmer: Reverted.

zea/data/spec.py
Felixduelmer: True, that simplifies this PR and is also in future easier to work with and understand. I have now changed it to accept both (n_frames, n_tx) or (n_frames*n_tx -1) and then directly in spec pad it with a zero. No changes needed anymore in `file.py`.

OisinNolan: Perfect, thanks @Felixduelmer ! 
I just notice now that the linter is failing -- this can be automatically corrected using `pre-commit run --all-files` ([docs ref](https://zea.readthedocs.io/en/latest/contributing.html#:~:text=pre%2Dcommit%20run%20%2D%2Dall%2Dfiles))

Then should be good to go :)

tristan-deep:
@Felixduelmer make sure to run `pre-commit run --all` to also run the linter :) (or `pre-commit install` and it will do before every commit so you won't have to think about it)

Felixduelmer:
> @Felixduelmer make sure to run `pre-commit run --all` to also run the linter :) (or `pre-commit install` and it will do before every commit so you won't have to think about it)

@tristan-deep, thanks, installed it.

zea/data/spec.py
Felixduelmer: Fixed the linter.

OisinNolan (approved):
Nice work, lgtm!

PR #427, opened 2026-06-11
Allow for sampling frequency or relative timestamps for the 1D signal

Felixduelmer:
This PR allows either saving the sampling frequency for 1D signals or saving relative timestamps for events in the 1D signal. That way, irregular timestamps are allowed. 

## Summary by CodeRabbit

* **New Features**
  * Signal instances now support explicit timestamp arrays for precise sample timing, providing an alternative to sampling frequency specification for improved temporal flexibility in signal construction.

* **Tests**
  * Added test coverage for timestamp validation, including length matching with samples, monotonic ordering verification, zero-start enforcement, and mutual exclusivity with sampling frequency parameter.


zea/data/spec.py
```diff
@@ -1457,22 +1457,26 @@ def __post_init__(self):
...
-    """Base class for additional signals with timing and sampling-frequency metadata.
+    """Base class for additional signals with timing metadata.
 
     Args:
         start_time_offset: Time offset in seconds between the first transmit event
             of the ultrasound acquisition and sample 0 of this data. Negative
             means this data starts before the first transmit event; positive
             means it starts after.
-        sampling_frequency: Sampling frequency in Hz for the additional signal samples.
+        sampling_frequency: Sampling frequency in Hz for uniformly sampled data.
```
OisinNolan, line 1467: Maybe we should mention in the docs that at least one but not both of these params is necessary

zea/data/spec.py
```diff
@@ -1457,22 +1457,26 @@ def __post_init__(self):
...
+        timestamps: Explicit sample timestamps in seconds of shape (T,), relative
+            to sample 0. Must start at 0.
     """
 
     start_time_offset: np.ndarray | float
-    sampling_frequency: np.ndarray | float
+    sampling_frequency: np.ndarray | float | None = field(default=None, kw_only=True)
+    timestamps: np.ndarray | None = field(default=None, kw_only=True)
```
OisinNolan, line 1476: I think the default values here can just be `None`

OisinNolan (commented):
Nice work, this looks good! Just left a couple of minor comments.

Not sure why the notebook tests are failing -- @tristan-deep have you seen this kind of error before? Doesn't look like anything to do with huggingface was edited by the PR -- same issue on the other one
```
E   ImportError: cannot import name 'RepoFile' from 'huggingface_hub' (/usr/local/lib/python3.12/site-packages/huggingface_hub/__init__.py). Did you mean: 'RepoUrl'?
```

tristan-deep:
> Nice work, this looks good! Just left a couple of minor comments.
> 
> Not sure why the notebook tests are failing -- @tristan-deep have you seen this kind of error before? Doesn't look like anything to do with huggingface was edited by the PR -- same issue on the other one
> 
> ```
> E   ImportError: cannot import name 'RepoFile' from 'huggingface_hub' (/usr/local/lib/python3.12/site-packages/huggingface_hub/__init__.py). Did you mean: 'RepoUrl'?
> ```

Local host runners have been a bit brittle in the past week also due to many runs and caching issues that might be a result of #393. I think my fixes in #420 resolved some issues but not all. Will have a look.

OisinNolan:
@tristan-deep thanks for taking a look. It's interesting that the same thing didn't happen on https://github.com/tue-bmd/zea/pull/428, will try re-running

tristan-deep:
> @tristan-deep thanks for taking a look. It's interesting that the same thing didn't happen on #428, will try re-running

Ah I remember now, forks dont build their own image (security reasons) but use the latest main image, see [here](https://github.com/tue-bmd/zea/blob/3748fe86376ee6bb9a213960bb4bf7f118ff32c8/.github/workflows/tests.yaml#L92). Usually this is okay, but since openh-rf has such a big diff with main currently it leads to some issues. Will see how to best handle this.  

For now don't worry about it and I'll make sure the fix will arrive here at some point.

tristan-deep:
> > @tristan-deep thanks for taking a look. It's interesting that the same thing didn't happen on #428, will try re-running
> 
> Ah I remember now, forks dont build their own image (security reasons) but use the latest main image, see [here](https://github.com/tue-bmd/zea/blob/3748fe86376ee6bb9a213960bb4bf7f118ff32c8/.github/workflows/tests.yaml#L92). Usually this is okay, but since openh-rf has such a big diff with main currently it leads to some issues. Will see how to best handle this.
> 
> For now don't worry about it and I'll make sure the fix will arrive here at some point.

@OisinNolan Fixed in #429

zea/data/spec.py
Felixduelmer: That kw_only marker is needed because Signal has subclasses, like Signal1D, that require fields. Without it, dataclasses would complain about required subclass fields coming after defaulted base fields.

OisinNolan (approved):
🚀

PR #438, opened 2026-06-15
Fix auto timestamping

swpenninga:
Removes the automatic acquisition date that is set on non-human subjects.


tristan-deep:
Nice! Looks good. Just need to update the test so it checks that no date was set instead.

tristan-deep (approved):
Super!

PR #443, opened 2026-06-17
Optional warnings, replacement of `scan.py` with `parameters.py` and removal of `generate_zea_dataset`

swpenninga:
Regarding #351, attempting to solve three tickboxes at once for efficiency.

Proposed changes: 

The warnings now have a flag to ignore them, such that your terminal does not get flooded when processing entire datasets. Additionally, as suggested, some metadata warnings have been removed entirely, even if you do not pass the ignore_warnings flag. (ecg, voice_narration, etc.) are rare and optional.

generate zea dataset is completely removed, and the legacy format is tested through picmus, which we generated with legacy and are hosting on hf.

scan.py is removed and parameters.py was created, to be in line with the new file format.

## Summary by CodeRabbit

* **New Features**
  * Added `ignore_warnings` parameter to `File.create()` to suppress warnings during file creation.

* **Bug Fixes**
  * Reduced warning noise by marking rarely-used metadata fields as optional without warnings when missing.

* **Breaking Changes**
  * Removed `generate_zea_dataset()` function; use alternative dataset creation methods.
  * Reorganized module structure; `Parameters` is now properly located in the parameters module.


docs/source/notebooks/data/zea_simulation_example.ipynb
```diff
@@ -78,7 +78,7 @@
     "from zea import init_device\n",
     "from zea.simulator import simulate_rf\n",
     "from zea.probes import Probe\n",
-    "from zea.scan import Parameters\n",
+    "from zea.parameters import Parameters\n",
```
tristan-deep:
```python
from zea import Parameters, Probe
```

Prefer main namespace in the examples

tristan-deep: In the codebase we do use full path for circular import errors etc.

tristan-deep: Fixed!

tristan-deep:
Nice @swpenninga! 
- I think the scan -> parameters rename was succesful
- Nice solution for removing `generate_zea_dataset`. I wonder at this point can the entire data_format.py be removed? 
- Regarding warnings:
I tested with some of the conversion examples in [openh-rf](https://github.com/open-h/OpenH-RF/tree/main/examples/templates). The `ignore_warnings=True` works well, and it even surfaced warnings for me that i didn't spot before (because of the many warnings). So that is positive. I think with the `ignore_warnings=False` we can still be more aggressive. @wesselvannierop maybe you can also have a look, wdyt? For instance, still getting warnings about metrics or spatial maps. We could all tag those with the rare tag that @swpenninga added?

Some examples:
- When subject is not set a warning gets printed, which i think is valid (you almost always know). But when you set subject, but don't add fat_percentage you also get a warning, which probably can be omitted. 
- Spatial maps that are not set shouldn't be warned.
- Metrics shouldn't be warned when not set.

tristan-deep:
_The big purge_

tristan-deep (dismissed):
Nice! Tested it with quite some examples. Added a bit more rares and also updated docs to be more complete.

tristan-deep (approved):

PR #447, opened 2026-06-18
Option to zero-pad instead of skip short videos in the dataloader

Louisvh:
When training models with longer time horizons, many samples may be too short to fill the whole time dimension, but could still be useful for training. This adds an option to the dataloader to pad instead of skip videos that are too short for the requested number of frames.

## Release Notes

* **New Features**
  * Added `pad_incomplete_blocks` option to the data loading pipeline. When enabled, incomplete HDF5 files are zero-padded to match the requested frame count instead of being skipped. Default behavior remains unchanged.

* **Tests**
  * Added regression test for padding behavior with incomplete data files.

### DIFF
--- a/tests/data/test_dataloader.py
+++ b/tests/data/test_dataloader.py
@@ -128,6 +128,37 @@ def test_h5_data_source(file_path, key, n_frames, insert_frame_axis, request):
         )
 
 
+def test_pad_incomplete_blocks(dummy_hdf5):
+    """Files shorter than a block are skipped by default and padded when enabled."""
+    n_frames = DUMMY_N_FRAMES + 50
+
+    skipped = H5DataSource(
+        file_paths=[dummy_hdf5],
+        key="data",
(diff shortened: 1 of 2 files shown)


swpenninga:
Just out of curiosity, what would you use this for? I dont think zero-ed out datapoints would help when training a model.

Louisvh:
For the 25-frame model, I wanted to include camus, but a lot of those videos are 17~20 frames. By zero-padding them, the network can hopefully still learn from the structure of the more recent frames. If I prepend black frames (with some random offset jitter) it introduces a bias towards black frames in the earliest history, but for posterior sampling that usually shouldn't matter. 

Any padding jitter logic can be done based on the returned indices; I figured more elaborate padding logic would be too specific bloat for the zea dataloader, so I put the user in charge of that for now.

swpenninga (approved):

PR #449, opened 2026-06-18
Update minimum version dependencies and test for version floor

### LINKED ISSUE #436: Carefully check dependency versions in `pyproject.toml`
Assuming the user does not always use `uv`, it could be that `pip` installs an older version of some dependency which is actually not compatible with `zea`.

wesselvannierop:
## Summary by CodeRabbit

* **Tests**
  * Added automated testing against minimum allowed dependency versions.
  * Improved test error diagnostics with detailed status reporting.

* **Chores**
  * Updated dependency version constraints for `matplotlib`, `grain`, and `jax` to ensure compatibility and stability.


tristan-deep:
Super! But now tests run twice, am I seeing that correct?

wesselvannierop:
> Super! But now tests run twice, am I seeing that correct?

Indeed, that might be a bit overkill but wanted to check if it worked. I already checked all tests locally, so I will update to only the `test_imports.py`, which should already surface a lot.

tristan-deep (approved):

PR #450, opened 2026-06-19
Improve support custom elements

wesselvannierop:
## Custom data elements

People may want to use the zea format but store some custom, unsupported data alongside it. This is now possible via the new `custom` key.

### Writing

Pass a list of `CustomElement` objects to `File.create`:

```python
import numpy as np
import zea
from zea.data.file import CustomElement

zea.File.create(
    path,
    data=data,
    scan=scan,
    probe=probe,
    custom=[
        CustomElement(
            name="lens_correction",
            data=np.float32(1.5),
            description="Scalar one-way delay offset.",
            unit="wavelengths",
        ),
        # Optionally nest under a (sub)group via `group_name`
        CustomElement(
            name="profile",
            data=profile_array,
            description="Per-element lens profile.",
            unit="-",
            group_name="lens/profiles",
        ),
    ],
)
```

Elements are validated and written to a dedicated `custom` group, reusing the standard spec dataset logic (scalar/string/compression handling).

### Reading

```python
with zea.File(path) as f:
    f.custom  # -> list[CustomElement]
```

### Backwards compatibility

In legacy files these were stored under `non_standard_elements` (a.k.a. `additional_elements`). Such files are detected automatically and still load through `f.custom`, so existing data keeps working.

### Notes

- `custom` is threaded through the full save path: `File.create`, `FileSpec`, `save_file`, and the file operations (`sum`, `resave`, `extract`, `compound_*`) now preserve custom elements.
- The Verasonics converter now writes its extra buffers (e.g. `lens_correction`, `verasonics_image_buffer`) via this `custom` mechanism instead of a post-hoc legacy write.


tristan-deep (changes requested):
Awesome! Only thing i can think of is to also update the docs. We have a dedicated section on custom fields and your PR description might be a good way to start. Should differentiate between custom fields and custom spatial maps etc. 

See [docs page](https://zea.readthedocs.io/en/openh-rf-latest/data-acquisition.html#custom-fields).

tristan-deep (approved):

PR #459, opened 2026-06-23
Refactor `set_transmits` into `find_transmits`

wesselvannierop:
You can now also do `idx = parameters.find_transmits("focused")` for other purposes, and it nicely refactors the logic

## Summary by CodeRabbit

* **Refactor**
  * Streamlined transmit selection handling so transmit indices are computed consistently and applied to the active configuration. `set_transmits()` now validates inputs more strictly (including when transmit count isn’t configured).
* **Bug Fixes**
  * Improved edge-case behavior for selection modes such as “center”, “focused”, “diverging”, and “plane”, including safer handling of empty results. For invalid/empty selections, the system now warns appropriately instead of failing unexpectedly.

### DIFF
--- a/zea/parameters.py
+++ b/zea/parameters.py
@@ -588,7 +588,7 @@ class Parameters(BaseParameters):
             value = self._params.get("focus_distances")
             if value is None:
                 raise ValueError("No focus distances provided, cannot select plane wave transmits")
-            return np.concatenate([np.where(value == 0)[0], np.where(np.isinf(value))[0]]).tolist()
+            return np.where((value == 0) | np.isinf(value))[0].tolist()
 
         # Handle integer - select evenly spaced transmits
         if isinstance(selection, (int, np.integer)):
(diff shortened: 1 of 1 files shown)


zea/parameters.py
```diff
@@ -513,75 +544,51 @@ def set_transmits(self, selection):
...
         Returns:
-            The current instance for method chaining.
+            The selected transmit indices.
 
         Raises:
             ValueError: If the selection is invalid or incompatible with the scan.
         """
         n_tx_total = self._params.get("n_tx")
         if n_tx_total is None:
-            if selection is None:
```
tristan-deep, line 523: Does this remove support for calling `zea.File.load_parameters()` for image data files (such as camus)?

wesselvannierop: `find_transmits` will fail for image-only, but the logic you mention is moved here: https://github.com/tue-bmd/zea/blob/2fa778a5f3b94c6636a533834994918a9257e6cf/zea/parameters.py#L516-L518

tristan-deep: ah perfect!

OisinNolan (approved):
Nice refactor, implementation looks good to me!

PR #460, opened 2026-06-23
Allow arbitrary leading dims in masks.py

wesselvannierop:
Adds relevant tests and allows more flexible usage of the functions in masks.py

## Release Notes

* **New Features**
  * Added `ZEA_TEST_DEVICE` to control which device tests run on.

* **Tests**
  * Expanded unit coverage for mask conversion utilities, including edge cases and multi-dimensional batch inputs.
  * Added parametrized tests validating `nonzero` behavior with fixed `size` and `fill_value`.

* **Bug Fixes**
  * Updated `nonzero` to consistently return fixed-size, padded results (when `size` is set) across backends.
  * Improved mask utilities to support arbitrary leading dimensions and enforce valid image-size divisibility.


zea/func/tensor.py
```diff
@@ -182,29 +182,58 @@ def boolean_mask(tensor, mask, size=None):
...
+    """
+    # Use ops.nonzero when size is not used
+    if size is None:
         return ops.nonzero(x)
 
+    # Directly use jnp.nonzero if JAX backend is used
+    if keras.backend.backend() == "jax":
+        import jax.numpy as jnp
+
+        return jnp.nonzero(x, size=size, fill_value=fill_value)
```
tristan-deep, line 211: Just checking jitting `keras.ops.nonzero` still does not work probably right?

tristan-deep (approved):

PR #465, opened 2026-06-26
Fix nested pipeline jit bug

wesselvannierop:
Found a small bug where a nested pipeline, e.g., `Map` or `PatchedGrid` inside a `Pipeline` would also be jitted when `jit_options="pipeline"`. This is inefficient and should be avoided.

Additionally fixes some other bugs related to operations using `jit_compile` or `jit_options` incorrectly.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved JIT behavior for pipelines and mapped operations, making nested execution more predictable.
  * Better validation errors now appear for invalid settings and unsupported inputs.
  * Scan conversion now clearly requires coordinates when run inside a compiled pipeline.
  * Band-pass filtering validation now behaves correctly in eager mode and remains safe during compiled execution.
* **Tests**
  * Added coverage for JIT scoping, non-jittable operations, and parameter/type validation.


zea/ops/pipeline.py
```diff
@@ -91,9 +91,6 @@ def __init__(
 
         self._pipeline_layers: List[Union[Operation, "Pipeline"]] = list(operations)
 
-        if jit_options not in ["pipeline", "ops", None]:
```
tristan-deep, line 94: why don't we validate anymore?

tristan-deep (approved):
Nice! One none-blocking comment. Also, @benluijten these fixes might be interesting for your ongoing experiments with jitting pipelines.

zea/ops/pipeline.py
wesselvannierop: This validation was duplicated and lives here now:
https://github.com/tue-bmd/zea/blob/f4e4b3a9c55630b5b1b741ec9f5d8537d68c2687/zea/ops/pipeline.py#L460-L461

tristan-deep: Ah thanks! I was trying to find it but couldn't, thanks for pointing it out.

tristan-deep (approved):

PR #469, opened 2026-06-29
Add bmi to zea file spec

tristan-deep:
Per request adding BMI to subject field of the zea data file spec.

### How to test
Try saving BMI during `zea.File.create` (test some value ranges).

## Summary by CodeRabbit

* **New Features**
  * Added an optional BMI field to subject metadata.
  * Documented the new `kg/m²` unit in the units reference.
* **Bug Fixes**
  * Added BMI validation: rejects non-finite and out-of-range values with errors.
  * Emits warnings for unusual (but non-fatal) BMI values outside the typical range.
* **Tests**
  * Expanded test coverage for BMI validation behavior and updated subject-field tests to include BMI.
* **Documentation**
  * Updated the subject metadata specification to include BMI.

### DIFF
--- a/docs/source/_spec_ref.rst
+++ b/docs/source/_spec_ref.rst
@@ -1028,6 +1028,11 @@ Fields marked :bdg-secondary:`optional` may be absent; all others are
               - scalar
               - Subject fat percentage.
               - |badge-opt|
+            * - ``bmi``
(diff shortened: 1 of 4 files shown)


swpenninga (approved):
Nice! i like the value verifications!

PR #470, opened 2026-06-29
Filter datasets by file content in `Dataset` / `Dataloader`

tristan-deep:
TLDR: Since #358 we have a lot of metadata available in the zea file format. The idea is that now you can can filter out files using `zea.Dataset` and `zea.Dataloader` using a simple filtering function.

## Summary

Adds a `file_filter` parameter to `zea.Dataset` and `zea.Dataloader` so you can load a
folder but keep only the files whose **content** matches a predicate.

Previously, file discovery was purely filesystem-based (walk a directory, match `.hdf5`/`.h5`).
With the rich per-file metadata now exposed through the spec system, there was no way to
select files by what's *inside* them. That is addressed in this PR.

The filtering logic lives in **one place** (`zea.Dataset`); `zea.Dataloader` just
forwards the parameter, so there's no duplicated logic.

## API

`file_filter` accepts **either** form:

1. **Callable** `Callable[[File], bool]` — receives the open `zea.File` (so it can read
   `f.metadata`, `f.scan`, `f.probe`, `f.n_frames`, …) and returns `True` to keep the file.
   Maximum flexibility: cross-field logic, ranges, arithmetic — anything.

2. **Declarative dotted-dict** `dict[str, Any]` — maps a dotted attribute path on the `File`
   to a condition. The condition is the `EXISTS` helper (field must be present), a plain
   value (equality), or a callable on the resolved value. All entries are ANDed together.
   Config-friendly for the common presence/equality cases.

**Semantics:** a file is kept iff the predicate returns `True`. If the predicate *raises*
(e.g. the file has no `metadata` group, so `f.metadata` raises `KeyError`), the file is
**excluded** with a debug log — so "keep files that have X" reads naturally without
defensive guards. A filter that removes everything raises a clear `ValueError`. Because
filtering must read each file, it is incompatible with `lazy=True` (raises with guidance).

```python
from zea import Dataloader, EXISTS

# callable form — full flexibility
Dataloader(
    file_paths="/data/ds",
    file_filter=lambda f: f.metadata.subject is not None
    and f.metadata.subject.fat_percentage is not None,
)

# dict form — presence + equality + value-level predicate (all ANDed)
Dataloader(
    file_paths="/data/ds",
    file_filter={
        "metadata.subject.fat_percentage": EXISTS,
        "metadata.subject.sex": "f",
        "scan.center_frequency": lambda v: 4e6 <= v <= 6e6,
    },
)
```

## Try it (standalone)

<details>
<summary>Runnable demo — writes two tiny zea files (one <em>with</em> a subject fat percentage,
one <em>without</em>), then shows the dict and callable filters keeping only the right file.
Self-cleaning. (click to expand)</summary>

```python
"""Standalone demo of Dataset/Dataloader content filtering. Run: python demo_filter.py"""

from pathlib import Path

import numpy as np

from zea import EXISTS, Dataset
from zea.data.file import File


def write_file(path, *, fat_percentage=None, center_frequency=5e6):
    """Write a minimal valid zea file with an `image` product and some metadata."""
    n_frames, n_tx, n_ax, n_el, n_ch, grid = 4, 2, 64, 8, 1, 16
    data = {
        "raw_data": np.ones((n_frames, n_tx, n_ax, n_el, n_ch), dtype=np.float32),
        "image": {"values": np.zeros((n_frames, grid, grid), dtype=np.uint8)},
    }
    scan = {
        "sampling_frequency": np.float32(40e6),
        "center_frequency": np.float32(center_frequency),
        "demodulation_frequency": np.float32(center_frequency),
        "initial_times": np.zeros((n_tx,), dtype=np.float32),
        "t0_delays": np.zeros((n_tx, n_el), dtype=np.float32),
        "tx_apodizations": np.ones((n_tx, n_el), dtype=np.float32),
        "focus_distances": np.full(n_tx, np.inf, dtype=np.float32),
        "transmit_origins": np.zeros((n_tx, 3), dtype=np.float32),
        "polar_angles": np.zeros((n_tx,), dtype=np.float32),
    }
    probe_geometry = np.zeros((n_el, 3), dtype=np.float32)
    probe_geometry[:, 0] = np.linspace(-0.02, 0.02, n_el)

    subject = {"sex": "f"}
    if fat_percentage is not None:
        subject["fat_percentage"] = np.float32(fat_percentage)

    File.create(
        path,
        data=data,
        scan=scan,
        probe={"name": "demo", "probe_geometry": probe_geometry},
        metadata={"subject": subject},
        description="filtering demo",
        overwrite=True,
    )

directory = "./demo-dataset"
directory = Path(directory)
directory.mkdir(exist_ok=True)

write_file(directory / "with_fat.hdf5", fat_percentage=17.5, center_frequency=5e6)
write_file(directory / "no_fat.hdf5", fat_percentage=None, center_frequency=9e6)

names = lambda ds: sorted(Path(p).name for p in ds.file_paths)

# no filter: both files
print("all files          :", names(Dataset(directory, validate=False)))

# dict filter: keep only files that record a subject fat percentage
ds = Dataset(directory, validate=False, file_filter={"metadata.subject.fat_percentage": EXISTS})
print("has fat_percentage :", names(ds))

# callable filter: keep only files in a center-frequency band
ds = Dataset(directory, validate=False, file_filter=lambda f: 4e6 <= float(f.scan.center_frequency) <= 6e6)
print("center freq 4-6MHz :", names(ds))
```

Expected output:

```
all files          : ['no_fat.hdf5', 'with_fat.hdf5']
has fat_percentage : ['with_fat.hdf5']
center freq 4-6MHz : ['with_fat.hdf5']
```

</details>

## Notes / trade-offs

- Filtering opens each candidate file to read its (small) metadata, so there's an up-front
  pass over all files. For `hf://` datasets that would mean downloading everything, which is
  why `file_filter` + `lazy=True` raises rather than silently downloading.

## Summary by CodeRabbit

* **New Features**
  * Added file filtering support when loading datasets and dataloaders, letting you include or exclude files with either a custom function or simple metadata rules.
  * Exposed a new `EXISTS` check for metadata-based filtering.

* **Bug Fixes**
  * Files without the required metadata are now skipped safely.
  * Filtering now fails clearly when it would remove every file, and it’s not allowed in incompatible loading modes.


swpenninga:
Very cool! I wonder, would this be a good PR to also create the multiple-dataset-dataloader? 
The main use-case of the filter would be to create data across multiple OpenH-RF submissions, right?

tristan-deep:
> Very cool! I wonder, would this be a good PR to also create the multiple-dataset-dataloader? The main use-case of the filter would be to create data across multiple OpenH-RF submissions, right?

Is this not already allowed? You can specify multiple folders (to both `zea.Dataset` and `zea.Dataloader` it should include in its search for `zea.Files`. Or is this not what you mean? We for sure should test this out with larger datasets etc. to see how it holds up :)

swpenninga:
> > Very cool! I wonder, would this be a good PR to also create the multiple-dataset-dataloader? The main use-case of the filter would be to create data across multiple OpenH-RF submissions, right?
> 
> Is this not already allowed? You can specify multiple folders (to both `zea.Dataset` and `zea.Dataloader` it should include in its search for `zea.Files`. Or is this not what you mean? We for sure should test this out with larger datasets etc. to see how it holds up :)

That is what I mean, but indeed, i was wondering if you tested multiple datasets with the new filter, and with hf paths rather than local storage?

tristan-deep:
> > > Very cool! I wonder, would this be a good PR to also create the multiple-dataset-dataloader? The main use-case of the filter would be to create data across multiple OpenH-RF submissions, right?
> > 
> > 
> > Is this not already allowed? You can specify multiple folders (to both `zea.Dataset` and `zea.Dataloader` it should include in its search for `zea.Files`. Or is this not what you mean? We for sure should test this out with larger datasets etc. to see how it holds up :)
> 
> That is what I mean, but indeed, i was wondering if you tested multiple datasets with the new filter, and with hf paths rather than local storage?

Yes this just works! But with a few caviats:

You can just pass a list of paths (for setup use the PR description stand-alone snippet):

```python
ds = Dataset([directory, "/data/2023_OPENH_RF_carotid"], validate=False, file_filter=lambda f: 5e6 <= float(f.scan.center_frequency) <= 8e6)
```

or with a Hugging Face path also allowed. 

```python
ds = Dataset([directory, "hf://zeahub/camus-sample"], validate=False, file_filter=lambda f: 4e6 <= float(f.scan.center_frequency) <= 6e6)
```

### Caviats
- For the filtering, it needs to be able to inspect the file quickly. This is fast, but needs the file locally available, so `zea.Dataset(..., lazy=True)` is not an option currently in combination with filtering.
- While `zea.Dataset` allows for combining arbitrary folders and paths (just pass a list of paths / folders), `zea.Dataloader` needs to be able to batch the data. If you don't have the right transforms in place, it might not be so straightforward. I think that is expected behavior, but we can offer more support for this. Such as transforms and documentation. 
- Currently the `zea.Dataloader` does not return any metadata, just the key that you are indexing. This might be a nice feature for the future.

swpenninga (approved):
Nice! I will approve the changes, but perhaps good to wait with merging for additional feedback from Discord?

tristan-deep:
> Nice! I will approve the changes, but perhaps good to wait with merging for additional feedback from Discord?

I think we can design this as we go and create follow-up PRs to add features etc. I will wait with a release until we have it dialed in a bit :)

PR #473, opened 2026-06-29
Improve parameter descriptions

tristan-deep:
Added expected parameter values for `focus_distances` and `tx_apodizations` to documentation.

## Summary by CodeRabbit

* **Documentation**
  * Clarified transmit timing parameter conventions in scan specifications, including apodization value interpretation within the expected `[-1, 1]` range.
  * Expanded focus-distance guidance: positive vs. negative finite values (focused vs. diverging) and plane-wave markers using `np.inf`, with `0` also accepted.
  * Improved transmit selection documentation by explicitly mapping “focused,” “diverging,” and “plane” behaviors to these conventions.

### DIFF
--- a/zea/data/spec.py
+++ b/zea/data/spec.py
@@ -1225,9 +1225,22 @@ class ScanSpec(Spec):
             full contribution. Negative values indicate that the element was
             fired with opposite polarity.
         focus_distances: The transmit focus distances in meters of shape (n_tx,).
-            This is the distance from the origin point on the transducer to
-            where the beam comes to focus. For planewaves this is set to
-            infinity or zero.
+            This is the distance from the transmit origin on the transducer to
+            where the beam comes to focus. The sign and magnitude encode the
+            transmit type:
+
+            - **positive finite**: focused transmit; the beam focuses at this
+              distance in front of the array.
+            - **negative finite**: diverging transmit; the (virtual) source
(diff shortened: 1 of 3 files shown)


zea/data/spec.py
```diff
@@ -1225,9 +1225,22 @@ class ScanSpec(Spec):
...
+            This is the distance from the transmit origin on the transducer to
+            where the beam comes to focus. The sign and magnitude encode the
+            transmit type:
+
+            - **positive finite**: focused transmit; the beam focuses at this
+              distance in front of the array.
+            - **negative finite**: diverging transmit; the (virtual) source
+              lies this distance behind the array.
+            - **infinite** (``np.inf``): plane wave. This is the preferred,
+              canonical value for plane waves in zea. ``0.0`` is also accepted
```
swpenninga, line 1237: i was always wondering what 0.0 actually did under the hood, and whether its a bad practice, or actually common practice?

swpenninga (approved):
Clear improvement

wesselvannierop (approved):

zea/data/spec.py
tristan-deep: Theoretically inf is the correct one, and zero is just a proxy that Verasonics uses as well. I’d say use inf whenever you can.

PR #475, opened 2026-06-30
Allow transmit-only tracks

OisinNolan:
This PR adds a relaxation to the `TrackSpec`, allowing us to store 'transmit-only' tracks, where we provide a scan but no data.

This could be useful for storing shear-wave elastography datasets (and perhaps other situations in future, e.g. from theranostics) where we want to save some the details of some transmit but don't care about recording the resulting echoes.

Note: we could consider adding extra validation here to make it so that:
(i) transmit-only tracks can only be specified in multi-track files 
(ii) a track_schedule and time_to_next transmit have to be specified if a transmit-only track is used
but I chose to leave the change quite minimal to avoid adding more complex validation logic. Curious what your thoughts are on that.

Example snippet:
```python
"""Create a zea file with B-mode and transmit-only push tracks via File.create."""

import numpy as np

from zea.data.file import File

n_frames, n_el, n_ax = 3, 4, 8

# B-mode: a handful of steered plane-wave transmits used to form an image.
bmode_n_tx = 3
bmode_angles = np.linspace(-5, 5, bmode_n_tx, dtype=np.float32)

# Push: a single, long, tightly-focused transmit (e.g. ARFI/shear-wave push).
push_n_tx = 1
push_focus_distance = np.float32(30e-3)


def bmode_scan():
    return {
        "sampling_frequency": np.float32(30e6),
        "center_frequency": np.float32(5e6),
        "demodulation_frequency": np.float32(5e6),
        "initial_times": np.zeros((bmode_n_tx,), dtype=np.float32),
        "t0_delays": np.zeros((bmode_n_tx, n_el), dtype=np.float32),
        "tx_apodizations": np.ones((bmode_n_tx, n_el), dtype=np.float32),
        "focus_distances": np.zeros((bmode_n_tx,), dtype=np.float32),
        "transmit_origins": np.zeros((bmode_n_tx, 3), dtype=np.float32),
        "polar_angles": np.deg2rad(bmode_angles),
    }


def push_scan():
    return {
        "sampling_frequency": np.float32(30e6),
        "center_frequency": np.float32(1.5e6),
        "demodulation_frequency": np.float32(1.5e6),
        "initial_times": np.zeros((push_n_tx,), dtype=np.float32),
        "t0_delays": np.zeros((push_n_tx, n_el), dtype=np.float32),
        "tx_apodizations": np.ones((push_n_tx, n_el), dtype=np.float32),
        "focus_distances": np.full((push_n_tx,), push_focus_distance, dtype=np.float32),
        "transmit_origins": np.zeros((push_n_tx, 3), dtype=np.float32),
        "polar_angles": np.zeros((push_n_tx,), dtype=np.float32),
    }


File.create(
    "tracks.hdf5",
    tracks=[
        # B-mode track: transmits and records channel data.
        {
            "data": {
                "raw_data": np.zeros((n_frames, bmode_n_tx, n_ax, n_el, 1), np.float32)
            },
            "scan": bmode_scan(),
            "label": "bmode",
        },
        # Push track: transmits only, records no data (data=None + transmit_only=True).
        {
            "data": None,
            "scan": push_scan(),
            "label": "push",
            "transmit_only": True,
        },
    ],
    probe={"probe_geometry": np.zeros((n_el, 3), dtype=np.float32)},
    overwrite=True,
)

with File("tracks.hdf5") as f:
    print("tracks:", f.track_labels)
    for track in f.tracks:
        print(f"  {track.label}: has data group =", "data" in track._group)
```

## Summary by CodeRabbit

* **New Features**
  * Added support for transmit-only tracks in file and track specifications.
  * Track metadata and scan information can now be preserved even when no data payload is present.

* **Bug Fixes**
  * Improved validation so empty track definitions are rejected unless transmit-only mode is explicitly set.
  * Saved files now round-trip transmit-only tracks correctly, including their scan details and labels.

* **Tests**
  * Added coverage for single-track and multi-track transmit-only file behavior.

### DIFF
--- a/tests/data/test_file.py
+++ b/tests/data/test_file.py
@@ -216,6 +216,7 @@ def test_create_transmit_only_track(tmp_path):
             {
                 "data": None,
                 "scan": _scan_minimal(n_frames=n_frames, n_tx=n_tx, n_el=n_el),
+                "transmit_only": True,
             }
         ],
         ignore_warnings=True,
@@ -246,6 +247,7 @@ def test_create_multitrack_with_transmit_only_track(tmp_path):
                 "data": None,
                 "scan": _scan_minimal(n_frames=n_frames, n_tx=n_tx, n_el=n_el),
                 "label": "push",
+                "transmit_only": True,
             },
         ],
         probe=_probe_minimal(n_el=n_el),
--- a/tests/data/test_file.py
+++ b/tests/data/test_file.py
@@ -237,9 +237,7 @@ def test_create_multitrack_with_transmit_only_track(tmp_path):
         tracks=[
             {
                 "data": {
-                    "raw_data": np.zeros(
-                        (n_frames, n_tx, n_ax, n_el, n_ch), dtype=np.float32
-                    ),
+                    "raw_data": np.zeros((n_frames, n_tx, n_ax, n_el, n_ch), dtype=np.float32),
                 },
                 "scan": _scan_minimal(n_frames=n_frames, n_tx=n_tx, n_el=n_el),
                 "label": "focused",
--- a/tests/data/test_file.py
+++ b/tests/data/test_file.py
@@ -206,6 +206,66 @@ def _probe_minimal(name=None, n_el=4):
     return probe
 
 
+def test_create_transmit_only_track(tmp_path):
+    """File.create with a single transmit-only track (scan but no data) works."""
+    n_frames, n_tx, n_el = 3, 2, 4
+    path = tmp_path / "transmit_only.hdf5"
+    File.create(
+        path,
+        tracks=[
+            {
+                "data": None,
+                "scan": _scan_minimal(n_frames=n_frames, n_tx=n_tx, n_el=n_el),
+            }
+        ],
+        ignore_warnings=True,
+    )
+
+    with File(path) as f:
+        assert isinstance(f.scan, ScanSpec), "f.scan should return a ScanSpec"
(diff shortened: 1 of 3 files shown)


tristan-deep:
Just checking, in the example you post in the PR description you pass the same scan to both tracks, but this would be a usecase for passing two different scan objects right?

OisinNolan:
@tristan-deep yeah good point -- generally speaking, we would expect something like a push pulse to have different scan parameters to a b-mode. I haven't added any validation to enforce that the scans are different, since I think that may be overkill, but if we add a docs example on this (e.g. based on the Oslo data) we would probably have different scans

tristan-deep:
> @tristan-deep yeah good point -- generally speaking, we would expect something like a push pulse to have different scan parameters to a b-mode. I haven't added any validation to enforce that the scans are different, since I think that may be overkill, but if we add a docs example on this (e.g. based on the Oslo data) we would probably have different scans

Okay makes sense and I agree! Maybe do update the PR description or docstrings just so people can refer to this and know how it should be used (two different scan classes). For me good to go!

OisinNolan:
@tristan-deep based on the discussion today I've added a `transmit_only` param to tracks to ensure that people are explicit about creating a track without data. This also gives us the opportunity to provide some more documentation + make the intended usage clearer, so I think it has improved the design.

wesselvannierop (approved):
Nice one!

PR #479, opened 2026-07-07
Adding instructions for using custom operations

tristan-deep:
See [documentation](https://zea--479.org.readthedocs.build/en/479/pipeline.html#custom-operations).

## Summary by CodeRabbit

* **Documentation**
  * Added a new “Custom operations” section covering how to define, register, and use user-defined operations in pipelines.
  * Documented YAML save/load behavior, including how operation identity is resolved based on registration naming.
  * Included a complete end-to-end example showing custom operation registration, pipeline creation, YAML round-trip, and cleanup.

### DIFF
--- a/docs/source/pipeline.rst
+++ b/docs/source/pipeline.rst
@@ -9,6 +9,121 @@ Pipeline
    :no-special-members:
    :no-index:
 
+.. _custom-ops:
+
+Custom operations
+-----------------------------------------
+
+You can define your own operations in your own project and use them everywhere a
+built-in ``zea`` operation works, including inside a :class:`~zea.ops.Pipeline` and saving them
+to a YAML config, without needing to change the ``zea`` source. In case you do want to add your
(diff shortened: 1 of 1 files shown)


docs/source/pipeline.rst
```diff
@@ -9,6 +9,129 @@ Pipeline
...
+
+       from zea import Pipeline
+       from zea.internal.registry import ops_registry
+       from zea.ops import Normalize, Operation
+
+
+       # Defined in this script, so register under "__main__". In a real project,
+       # register under your module path (e.g. "my_project.my_ops.MyScale") so the
+       # saved YAML can be loaded from anywhere, as described above.
+       @ops_registry("__main__.MyScale")
```
wesselvannierop, line 111: Maybe use `my_scale` like above?

tristan-deep: If you define the path like that it is a bit more robust as it will find the module this custom operation is defined in. That has more value than a custom name like `my_scale` where you should import the custom operation first.

PR #481, opened 2026-07-08
Fix transmit-delay discontinuity at the focus

tristan-deep:
## What changed
- Added `focal_region_length`: a focal-region delay-smoothing parameter for focused transmits (defined as full length around the focal plane).
- Changed pressure-field resize interpolation default to `bilinear` (override remains available via `pfield_kwargs["interpolation"]`).

See here with `focal_region_length` on (`2e-3`) and off. The difference is subtle but for this carotid example noticeable.

comparison_all

<details>
<summary>Example</summary>

```python

import os

os.environ.setdefault("KERAS_BACKEND", "jax")

import keras
import numpy as np

import zea

zea.init_device()

# ---------------------------------------------------------------------------
# 1. In vivo example: beamform a real carotid acquisition before/after the fix.
# ---------------------------------------------------------------------------
# can change to your own path (local or huggingface) of matching dataset file and config
path = "hf://zeahub/zea-carotid-2023"
config_path = "hf://zeahub/zea-carotid-2023/config.yaml"

config = zea.Config.from_path(config_path)
config.parameters.dynamic_range = (-60, 0)
pipeline = zea.Pipeline.from_config(config)

with zea.Dataset(path, lazy=True) as dataset:
    file = dataset[2]
    parameters = file.load_parameters()
    parameters.update(config.parameters)
    raw = file.data.raw_data[:1][:, parameters.selected_transmits]

# focal_region_length=0.0 (default) -> conventional hard switch; 2e-3 -> blended focal region.
invivo_images = []
for length in (0.0, 2e-3):
    inputs = pipeline.prepare_parameters(parameters, focal_region_length=length)
    image = pipeline(**{pipeline.key: raw, **inputs})[pipeline.output_key]
    invivo_images.append(keras.ops.squeeze(image))

# ---------------------------------------------------------------------------
# 2. Point-target example: same before/after, but on a simulated focused
#    transmit so the focal-region artefact is unambiguous.
# ---------------------------------------------------------------------------
from zea import Parameters, Probe
from zea.beamform import phantoms
from zea.beamform.delays import compute_t0_delays_focused
from zea.simulator import simulate_rf

n_el, n_tx = 64, 1
sound_speed = 1540.0
# Pitch set by a 3 MHz wavelength (wide ~33 mm aperture) → tight focus and
# pronounced focal-region sidelobes, even though we image at 5 MHz.
pitch = sound_speed / 3e6
probe = Probe(
    probe_geometry=np.stack(
        [(np.arange(n_el) - (n_el - 1) / 2) * pitch, np.zeros(n_el), np.zeros(n_el)],
        axis=1,
    ),
    probe_center_frequency=5e6,
)
transmit_origins = np.zeros((n_tx, 3))
focus_distances = np.ones(n_tx) * 20e-3
polar_angles = np.zeros(n_tx)
xlims, zlims = (-20e-3, 20e-3), (10e-3, 35e-3)
wavelength = sound_speed / probe.probe_center_frequency

sim_parameters = Parameters(
    n_tx=n_tx,
    n_el=n_el,
    center_frequency=probe.probe_center_frequency,
    sampling_frequency=probe.probe_center_frequency * 4,
    t0_delays=compute_t0_delays_focused(
        transmit_origins=transmit_origins,
        focus_distances=focus_distances,
        polar_angles=polar_angles,
        sound_speed=sound_speed,
    ),
    tx_apodizations=np.ones((n_tx, n_el)) * np.hanning(n_el)[None],
    element_width=float(np.linalg.norm(probe.probe_geometry[1] - probe.probe_geometry[0])),
    focus_distances=focus_distances,
    polar_angles=polar_angles,
    transmit_origins=transmit_origins,
    initial_times=np.ones(n_tx) * 1e-6,
    n_ax=1024,
    n_ch=1,
    xlims=xlims,
    zlims=zlims,
    grid_size_x=int((xlims[1] - xlims[0]) / (0.5 * wavelength)) + 1,
    grid_size_z=int((zlims[1] - zlims[0]) / (0.5 * wavelength)) + 1,
    sound_speed=sound_speed,
    apply_lens_correction=False,
    lens_sound_speed=1000,
    lens_thickness=1e-3,
    attenuation_coef=0.0,
    selected_transmits="all",
)

# Fish phantom straddling the focal plane (focus at 20 mm): its sidelobes sit
# right around the focus, so the focal-region artefact is clearly visible.
positions = phantoms.fish() + np.array([-8e-3, 0, 0])[None]
rf_data = simulate_rf(
    scatterer_positions=positions,
    scatterer_magnitudes=np.ones(len(positions), dtype=np.float32),
    apply_lens_correction=sim_parameters.apply_lens_correction,
    lens_thickness=sim_parameters.lens_thickness,
    lens_sound_speed=sim_parameters.lens_sound_speed,
    sound_speed=sim_parameters.sound_speed,
    n_ax=sim_parameters.n_ax,
    center_frequency=probe.probe_center_frequency,
    sampling_frequency=sim_parameters.sampling_frequency,
    t0_delays=sim_parameters.t0_delays,
    initial_times=sim_parameters.initial_times,
    element_width=sim_parameters.element_width,
    attenuation_coef=sim_parameters.attenuation_coef,
    tx_apodizations=sim_parameters.tx_apodizations,
)

sim_pipeline = zea.Pipeline.from_default(enable_pfield=False, with_batch_dim=False, baseband=False)
point_images = []
for length in (0.0, 10e-3):
    inputs = sim_pipeline.prepare_parameters(sim_parameters, focal_region_length=length)
    image = sim_pipeline(**{sim_pipeline.key: rf_data, **inputs})[sim_pipeline.output_key]
    point_images.append(keras.ops.squeeze(image))

# ---------------------------------------------------------------------------
# 3. Plot: rows = in vivo / point target, columns = before / after / difference.
# ---------------------------------------------------------------------------

zea.visualize.set_mpl_style()

rows = [("In vivo", invivo_images, (-60, 0)), ("Point target", point_images, (-50, 0))]
fig, axes = plt.subplots(2, 3, figsize=(12, 8))

for row, (label, images, dynamic_range) in enumerate(rows):
    img0 = zea.display.to_8bit(images[0], dynamic_range=dynamic_range)
    img1 = zea.display.to_8bit(images[1], dynamic_range=dynamic_range)
    difference = np.array(images[1]) - np.array(images[0])

    axes[row, 0].imshow(img0, cmap="gray")
    axes[row, 1].imshow(img1, cmap="gray")

    abs_diff = np.abs(difference.flatten())
    abs_diff = abs_diff[~np.isnan(abs_diff)]
    vmax = np.percentile(abs_diff, 99) if len(abs_diff) > 0 else 1.0
    im = axes[row, 2].imshow(difference, cmap="coolwarm", vmin=-vmax, vmax=vmax)
    plt.colorbar(im, ax=axes[row, 2], fraction=0.046, pad=0.04)

    axes[row, 0].set_ylabel(label)
    for ax in axes[row]:
        ax.set_xticks([])
        ax.set_yticks([])

for col, title in enumerate(("Before", "After", "Difference")):
    axes[0, col].set_title(title)

plt.savefig("comparison_all.png", dpi=150, bbox_inches="tight")
print("saved comparison_all.png")
plt.close()
```

</details>

## Long story

For focused transmits, `transmit_delays` uses the **first-arrival** time for pixels before the focus and the **last-arrival** time for pixels beyond it. The hard switch at the focal plane introduces a kink in the modeled wavefront (a delay discontinuity). After delay-and-sum this appears as a sharp discontinuity, most visible in phantom / point-target recordings of focused-imaging (FI) datasets.

This is the virtual-source artifact described in [paper](https://doi.org/10.1109/ULTSYM.2018.8579944).

zea computes the transmit delay as the **first-arrival** time (min over elements) before the focus and the **last-arrival** time (max over elements) after it; the hard switch at the focal plane is the discontinuity.

Now, inside a small region around the focal plane we instead linearly blend the first- and last-arrival times (weight 0 → 1 across the band), which equals the surrounding values at the two edges and passes smoothly through the focus, so it removes the discontinuity while leaving the delays elsewhere unchanged. This implementation uses the same min/max arrival-time model that is already used outside the focal region. See:

https://github.com/tue-bmd/zea/blob/180af8b8e1b2b7f37da5327d66b14c87eb5603f0/zea/beamform/beamformer.py#L731-L737

The region parameter is `focal_region_length` (meters), defined as the full length centered at the focal plane (internally, +/- `focal_region_length / 2`). It only affects focused transmits (finite, positive `focus_distance`); plane-wave and diverging-wave transmits are untouched. The default is `0.0` (disabled, i.e. the current behavior), so this is opt-in and fully backward-compatible.

### Pressure field weighting interpolation
Pressure-field resize interpolation now defaults to `bilinear` (to reduce blocky transitions from the downsampled pressure field). `nearest` remains available as an override. See comparison below. Note that with pressure field weighting (RTB) turned on, the above `focal_region_length` effect becomes small in this example.

```python
# default behavior now (bilinear)
config.parameters.pfield_kwargs = {
    "alpha": 30.0,
    "percentile": 10,
}

# optional override for comparison
for interp_method in ["nearest", "bilinear"]:
    config.parameters.pfield_kwargs = {
        "interpolation": interp_method,
        "alpha": 30.0,
        "percentile": 10,
    }
    parameters.update(config.parameters)
    ...
```

pfield_interpolation_comparison

## Summary by CodeRabbit

* **New Features**
  * Added optional `focal_region_length` to focused transmit delay calculations to enable hybrid first/last delay blending around the focal plane (disabled by `0`/`None`).
  * Added `interpolation` selection for pressure-field resizing (default: bilinear).

* **Bug Fixes**
  * Reduced transmit-delay discontinuities near focus when blending is enabled.
  * Added warnings when `focal_region_length` is provided but would not be used for the selected focus mode.

* **Tests**
  * Added unit coverage for `focal_region_length` behavior.

* **Refactor**
  * Minor internal change to parameter preparation.


rvsloun:
Great fix! Seems that it is not equivalent to Rindal et al., who blend a plane wave and spherical virtual source model around the focus, although practically it may be very close. Perhaps good to clarify that it is not an exact implementation of Rindal et al., as was somehow suggested by the comment? Perhaps also good to compare against Rindal's implementation to check if there is a practical difference.

zea/beamform/pfield.py
```diff
@@ -50,6 +50,7 @@ def compute_pfield(
     percentile=10,
     norm=True,
     point_batch_size=2048,
+    interpolation="nearest",
```
wesselvannierop: I'd be up for 'bilinear' as default, makes more sense.

tristan-deep: How much slower is it? Was the only consideration I had.

wesselvannierop: And increasing the pressure field resolution would also makes sense now it is much faster. I always do that myself now.

wesselvannierop: I can quickly check for speed, but I wouldn't be too concerned "bilinear" is usually the default for resizing.

wesselvannierop:
> Great fix! Seems that it is not equivalent to Rindal et al., who blend a plane wave and spherical virtual source model around the focus, although practically it may be very close. Perhaps good to clarify that it is not an exact implementation of Rindal et al., as was somehow suggested by the comment? Perhaps also good to compare against Rindal's implementation to check if there is a practical difference.

I'll have a look, I looked into this a while ago anyway.

zea/beamform/pfield.py
tristan-deep: makes sense! feel free to change to the more appropriate defaults (and check with snippet above if it produces ok results?)

zea/beamform/beamformer.py
```diff
@@ -374,6 +406,11 @@ def calculate_delays(
             m/s.
         n_iter (int, optional): Newton-Raphson iterations for lens
             correction.  Defaults to ``2``.
+        focal_region_margin (float, optional): Half-width in meters of the
```
vincentvdschaft: I assumed that margin meant the whole width instead of the half width. Perhaps this naming could be more clear. I would suggest either having half in the name, or just having the parameter be the length. Something like `focal_region_half_length`, `focal_region_length`? What do you think?

vincentvdschaft:
It seems to work well.
simulation_plot_images

I do think there is an issue with the beamformer currently, but this seems unrelated to the current PR. Note the focal region border in the wrong location here:
image

I think the issue is here. The transmit origins are not taken into account for the pixel positions relative to the focus.

https://github.com/tue-bmd/zea/blob/3ef1ce1c6598611da1b266dfdec01155e0658ff8/zea/beamform/beamformer.py#L654-L662

Do we want to fix this in here or shall I open a new PR for this?

tristan-deep:
> Do we want to fix this in here or shall I open a new PR for this?

@vincentvdschaft thanks for taking a look! Maybe a new PR makes sense? I think @wesselvannierop is working on this branch now so his call. 

Also could you post the snippet to obtain the comparison of the point target as you showed?

zea/beamform/pfield.py
wesselvannierop: 'bilinear' is 0.5 sec slower in tensorflow generally, but I think it is worth it because it is a one-time cost per transmit scheme. For jax no differences are observed.

vincentvdschaft:
> > Do we want to fix this in here or shall I open a new PR for this?
> 
> @vincentvdschaft thanks for taking a look! Maybe a new PR makes sense? I think @wesselvannierop is working on this branch now so his call.
> 
> Also could you post the snippet to obtain the comparison of the point target as you showed?

I double checked. It was actually my mistake. I think it works correctly after all.

wesselvannierop:
Current state of the code is a mix between the USTB 'hybrid' method and 'blended', so we might want to decide on which version we want. Or we keep this version, but this I think is not directly related to Rindal et al. and/or the USTB implementation.

tristan-deep:
> I double checked. It was actually my mistake. I think it works correctly after all.

@vincentvdschaft Great! Could you still post the snippet for the first point scatter image related to this PR?

vincentvdschaft:
> ! Could you still post the snippet for the first point scatter image related to this PR?

Sure. Here it is.

<details>
<summary>Code for generating the image</summary>
```python
import os

import numpy as np

import zea
from zea import Parameters, Probe, init_device
from zea.beamform import phantoms
from zea.beamform.delays import compute_t0_delays_focused
from zea.simulator import simulate_rf
from zea.visualize import set_mpl_style

# In[4]:


init_device(verbose=False)
set_mpl_style()


center_frequency = 3e6
n_el = 64
n_tx = 1
wavelength = 1540.0 / center_frequency
probe_geometry = np.stack(
    [
        (np.arange(n_el) - (n_el - 1) / 2) * wavelength,
        np.zeros(n_el),
        np.zeros(n_el),
    ],
    axis=1,
)
transmit_origins = np.array([0, 0, 0])[None] * np.ones((n_tx, 1))
focus_distances = np.ones(n_tx) * 20e-3
angles = np.array([0]) * np.pi / 180


def get_image(focal_region_margin=0.0):
    # Define a linear probe

    probe = Probe(
        probe_geometry=probe_geometry,
        probe_center_frequency=5e6,
    )

    # Define a planewave scan

    # 40 degrees in radians
    sound_speed = 1540.0

    # Set grid and image size
    xlims = (-20e-3, 20e-3)
    zlims = (10e-3, 35e-3)
    width, height = xlims[1] - xlims[0], zlims[1] - zlims[0]
    wavelength = sound_speed / probe.probe_center_frequency
    grid_size_x = int(width / (0.5 * wavelength)) + 1
    grid_size_z = int(height / (0.5 * wavelength)) + 1

    t0_delays = compute_t0_delays_focused(
        transmit_origins=transmit_origins,
        focus_distances=focus_distances,
        probe_geometry=probe_geometry,
        polar_angles=angles,
        sound_speed=sound_speed,
    )
    tx_apodizations = np.ones((n_tx, n_el)) * np.hanning(n_el)[None]

    parameters = Parameters(
        n_tx=n_tx,
        n_el=n_el,
        center_frequency=probe.probe_center_frequency,
        sampling_frequency=probe.probe_center_frequency * 4,
        probe_geometry=probe_geometry,
        t0_delays=t0_delays,
        tx_apodizations=tx_apodizations,
        element_width=np.linalg.norm(probe_geometry[1] - probe_geometry[0]),
        focus_distances=focus_distances,
        polar_angles=angles,
        initial_times=np.ones(n_tx) * 1e-6,
        n_ax=1024,
        xlims=xlims,
        zlims=zlims,
        grid_size_x=grid_size_x,
        grid_size_z=grid_size_z,
        lens_sound_speed=1000,
        lens_thickness=1e-3,
        n_ch=1,
        selected_transmits="all",
        sound_speed=sound_speed,
        apply_lens_correction=False,
        attenuation_coef=0.0,
        focal_region_margin=focal_region_margin,
        transmit_origins=transmit_origins,
    )

    # Create the phantom scatterer positions and magnitudes
    positions = phantoms.fish() + np.array([-8e-3, 0, 0])[None]
    magnitudes = np.ones(len(positions), dtype=np.float32)

    simulation_args = {
        "scatterer_positions": positions,
        "scatterer_magnitudes": magnitudes,
        "probe_geometry": probe.probe_geometry,
        "apply_lens_correction": parameters.apply_lens_correction,
        "lens_thickness": parameters.lens_thickness,
        "lens_sound_speed": parameters.lens_sound_speed,
        "sound_speed": parameters.sound_speed,
        "n_ax": parameters.n_ax,
        "center_frequency": probe.probe_center_frequency,
        "sampling_frequency": parameters.sampling_frequency,
        "t0_delays": parameters.t0_delays,
        "initial_times": parameters.initial_times,
        "element_width": parameters.element_width,
        "attenuation_coef": parameters.attenuation_coef,
        "tx_apodizations": parameters.tx_apodizations,
    }

    rf_data = simulate_rf(**simulation_args)
    print("Simulated RF data shape:", rf_data.shape)

    pipeline = zea.Pipeline.from_default(enable_pfield=False, with_batch_dim=False, baseband=False)
    inputs = pipeline.prepare_parameters(parameters, dynamic_range=(-50, 0))
    inputs[pipeline.key] = rf_data
    outputs = pipeline(**inputs)
    image = outputs[pipeline.output_key]

    image = zea.display.to_8bit(image, dynamic_range=(-50, 0))

    return image


fig, axes = plt.subplots(1, 2, figsize=(12, 5))
for i, focal_region_margin in enumerate([0.0, 5e-3]):
    image = get_image(focal_region_margin=focal_region_margin)
    axes[i].imshow(image, cmap="gray", extent=[-20, 20, 35, 10])
    axes[i].set_title(f"Focal Region Margin: {focal_region_margin * 1e3:.1f} mm")
    axes[i].set_xlabel("Lateral (mm)")
    axes[i].plot(
        probe_geometry[:, 0] * 1e3,
        np.zeros(probe_geometry.shape[0]),
        "C0-",
        linewidth=2,
        label="Probe Elements",
    )

axes[1].set_yticks([])
axes[0].set_ylabel("Axial (mm)")
# axes[1].legend(loc="lower right", fontsize=8)
plt.tight_layout()
plt.savefig("simulation_plot_images.png")
```
</details>

zea/beamform/pfield.py
tristan-deep: Made bilinear default now

zea/beamform/beamformer.py
tristan-deep: Renamed to `focal_region_length` which is now the full slab length.

tristan-deep:
> Great fix! Seems that it is not equivalent to Rindal et al., who blend a plane wave and spherical virtual source model around the focus, although practically it may be very close. Perhaps good to clarify that it is not an exact implementation of Rindal et al., as was somehow suggested by the comment? Perhaps also good to compare against Rindal's implementation to check if there is a practical difference.

Yes you are right, the artefact is described in the paper, but the solution we settled on is a bit different and in the end was a better fit for our implementation. I updated the erroneous docstrings.

tristan-deep:
@coderabbitai full review

zea/ops/pipeline.py
```diff
@@ -853,7 +853,23 @@ def prepare_parameters(
...
+        """Check the prepared parameters for potential issues."""
+
+        # Lightweight, pre-jit sanity check: flag a focal_region_length that
+        # cannot take effect because the data has no focused transmits.
+        if prepared.get("focal_region_length") is not None and "focus_distances" in prepared:
+            from zea.beamform.beamformer import warn_if_focal_region_length_unused
+
+            warn_if_focal_region_length_unused(
+                prepared["focus_distances"], prepared["focal_region_length"]
+            )
```
wesselvannierop: I understand that this check is nice, but I'm not sure `Pipeline` is the place. I think when a specific pipeline is being traced with the initial arguments, you could still just do this check inside the `tof_correction` function.

tristan-deep: I agree found this to be an awkward place, will move it.

tristan-deep: Done

tristan-deep:
Just updated the PR description with code snippet and new figure!

PR #474, opened 2026-06-30
Implemented us4us to zea converter

pjarosik:
Resolves https://github.com/tue-bmd/zea/issues/448

NOTE:
- Different types of data from the same acquisition (e.g., `raw_data` and `beamformed_data`) are stored in separate tracks. Otherwise, I encountered data consistency issues—for example, when `raw_data` contained raw RF data (rather than I/Q), while `beamformed_data` already contained I/Q data.
- In the tests, I used the dataset hosted on Dropbox. You may want to consider moving it to one of the repositories you already use (https://www.dropbox.com/scl/fi/xhid5kosgut2vzeyr5j19/zea_us4us_converter_test_data.pkl?rlkey=94e533yaebs2duijtk1mqnm79&dl=1)

## Summary by CodeRabbit

* **New Features**
  * Added CLI support to convert **us4us** recordings into ZEA **.hdf5** format, for both single-file and directory (batch) inputs.
  * Added `--mapping` (JSON, default `{"0":"image"}`) to generate one or multiple output tracks, plus `--overwrite` to replace existing outputs.
* **Documentation**
  * Replaced placeholder guidance with detailed us4us conversion instructions, including input/output options and `--mapping` examples.
* **Bug Fixes**
  * Hardened conversion against unsafe pickle contents and added stricter validation with clearer error reporting for malformed inputs.
* **Tests**
  * Extended end-to-end conversion tests for **us4us** using a cached, checksum-verified fixture and enhanced output shape/content checks (including multi-track mapping).

PR #483, opened 2026-07-08
Fixes for shape consistency across data types and tracks

wesselvannierop:
Allow n_ch to be globally different but locally consistent in a zea file. This arose in #474, where IQ and RF data where stored in the same file, which should be allowed.

Also fixes that the shapes in "annotations" must match at least one track, not all (since tracks can have different shapes in general)

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved dimension validation: channel-related counts are now checked only within the appropriate data product/map scope, while non-channel dimensions still remain consistent across the overall specification.
  * Allowed differing spatial channel counts across different map types within the same spec.
  * Kept raw vs beamformed channel counts independent, with errors still raised for mismatches within a single map where required.

* **Tests**
  * Extended unit coverage for the updated dimension-consistency and multi-track annotation validation rules.

* **Documentation**
  * Updated accepted annotation `view`/`label` inputs to allow either arrays or strings.

### DIFF
--- a/tests/data/test_spec.py
+++ b/tests/data/test_spec.py
@@ -1131,15 +1131,43 @@ class TestMetadataAndMetricsValidationErrors:
                 },
             )
 
-    def test_annotations_n_frames_mismatch_against_later_track_raises(self):
-        """Metadata may agree with track 0 but conflict with a later track.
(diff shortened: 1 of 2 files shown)


swpenninga (approved):
Nice!

PR #484, opened 2026-07-09
Do not cast `None` in spec / parameters

wesselvannierop:
This bug occurred when a data file contained `azimuth_angles=None` and when this gets cast in the `Parameter` class (to float32) it becomes `np.nan`. This fixes that.


swpenninga (approved):
I agree with the change, but is 18 tests not a bit wild?

PR #489, opened 2026-07-10
Speed and memory optimization of the beamforming pipeline

Louisvh:
This implements some low-hanging fruit optimizations in the beamforming pipeline. The biggest gains are in DMAS, which was re-computing a massive matrix of unused values for each pixel, which is now just computing the necessary values.

I've preserved the low-memory option on a separate branch, in case we want to revisit that later. It's removed from this PR.

Some quick stats on the bdmserver (2080ti, `jax` backend), using the `s1` image from `zeahub/zea-cardiac-2026`:

| grid  | version | DAS ms/frame | DAS peak mem | DMAS ms/frame | DMAS peak mem |
|----------|------------|---------:|---------:|---------:|---------:|
| _1024x2048_         |
| | v0.1.2          |    1343.2 | 7.22 GiB |   10997.4 | 4.03 GiB |
| | 5832b64 |    1096.5 | 3.34 GiB |    1652.3 | 3.34 GiB |
| | current |    639.5 | 2.11 GiB |    1014.2 | 2.65 GiB |
| _512x768_         |
| | v0.1.2          |     297.4 | 0.85 GiB |     2059.1 | 0.83 GiB |
| | 5832b64  |     224.3 | 0.70 GiB |      302.3 | 0.70 GiB |
| | current  |     133.0 | 0.47 GiB |      202.2 | 0.58 GiB |
| _400x566_         |
| | v0.1.2          |     160.6 | 0.48 GiB |   1147.7 | 0.48 GiB |
| | 5832b64  |     132.8 | 0.39 GiB |    166.6 | 0.41 GiB |
| | current  |     82.4 | 0.38 GiB |    114.9 | 0.38 GiB |
| _↑, pfield off_         |
| | v0.1.2          |     158.2 | 0.41 GiB |   1143.7 | 0.40 GiB |
| | 5832b64  |     132.3 | 0.33 GiB |    165.7 | 0.33 GiB |
| | current  |     80.8 | 0.26 GiB |    113.4 | 0.26 GiB |

## Summary by CodeRabbit

* **Performance**
  * Improved IQ phase correction efficiency during beamforming.
  * Streamlined delay interpolation and image reconstruction calculations.

* **Bug Fixes**
  * Improved handling of integer scan bounds when detecting cone parameters.
  * Maintained consistent sample clipping and interpolation behavior.

* **Refactor**
  * Updated complex rotation support for more efficient calculations.
  * Simplified DMAS beamforming computations without changing the resulting output.


wesselvannierop:
Very cool @Louisvh! So you mention the low memory version does `n_tx` summing sequentially. I have some time ago used this to reduce memory usage:

```python
batch_size_tx = 2
pixel_chunks = 10

beamformer = zea.ops.Map(
    [
        zea.ops.Map(
            [
                zea.ops.TOFCorrection(),
                zea.ops.PfieldWeighting(),
                zea.ops.keras_ops.Sum(axis=-2),  # sum over elements
            ],
            argnames=(
                # default:
                "data",  # make sure self.key is "data"
                "t0_delays",
                "tx_apodizations",
                "initial_times",
                "polar_angles",
                "focus_distances",
                "transmit_origins",
                "flat_pfield",
                "t_peak",
                # custom:
                "rx_apo",
            ),
            batch_size=batch_size_tx,  # only 2 tx at a time
            in_axes=(0, 0, 0, 0, 0, 0, 0, 1, 0, 0),
        ),
        zea.ops.keras_ops.Sum(axis=-3),  # sum over transmits
    ],
    chunks=pixel_chunks,
    argnames=("flatgrid", "flat_pfield", "rx_apo"),
    in_axes=(0, 0, 1),
)
```

How similar would this be to your implementation? I guess this one still does the summing afterwards, but maybe we can make a `zea.ops.Scan` which is able to achieve your memory reduction, with the simplicity of just defining it as a special `Pipeline`?

Louisvh:
Yes, that's very similar. The fused version is about 30% faster, which probably comes down to JIT not being as efficient across function boundaries. If we can make the pipeline-level version a bit more user friendly (maybe a convenience function that does the wrapping?), that's probably preferable over complicating the beamforming code (as in either case, it's explicitly a "slow" path aimed at memory savings).

wesselvannierop:
I'm assuming the `jit_options="pipeline"` should be as fast as the fused version? Or not?

Louisvh:
In a perfect world it would be, but the jit compiler apparently isn't that smart. When I set `jit_options="pipeline"`, none of the versions become faster, and peak memory goes up:

method | jit | ms/frame | peak mem
-----|----|----|----
default | ops | 1651.8 | 3.25 GiB
default | pipeline | 1650.9 | 3.35 GiB
fused | ops | 1793.4 | 1.17 GiB
fused | pipeline | 1790.0 | 1.58 GiB
map | ops | 3382.9 | 1.81 GiB
map | pipeline | 3383.4 | 2.14 GiB

wesselvannierop:
> In a perfect world it would be, but the jit compiler apparently isn't that smart. When I set `jit_options="pipeline"`, none of the versions become faster, and peak memory goes up:
> 
> method	jit	ms/frame	peak mem
> default	ops	1651.8	3.25 GiB
> default	pipeline	1650.9	3.35 GiB
> fused	ops	1793.4	1.17 GiB
> fused	pipeline	1790.0	1.58 GiB
> map	ops	3382.9	1.81 GiB
> map	pipeline	3383.4	2.14 GiB

Very interesting finding. We should investigate this, since it is not what I have experienced in the past. For me, moving more code into the `jit` context, always meant it was al lot faster. (that is using jax as backend)

Louisvh:
> Very interesting finding. We should investigate this, since it is not what I have experienced in the past. For me, moving more code into the `jit` context, always meant it was al lot faster. (that is using jax as backend)

I did some more digging, and I think that was from before the patchedgrid change. The Map is jitted as one block and pipeline jit won't push any other statements into or out of that loop, so ops-wise and pipeline-wise compilation are pretty similar.

While digging I did find another interesting thing though: for the IQ case, a lot of time in apply_delays was being spent in a data-shuffling operation that wasn't actually necessary. I've pushed a fix; the difference is even bigger now:

| grid  | version | DAS ms/frame | DAS peak mem | DMAS ms/frame | DMAS peak mem |
|----------|------------|---------:|---------:|---------:|---------:|
|  _400x566_ | v0.1.2          |     160.6 | 0.48 GiB | 1147.7 | 0.48 GiB |
| | Before indexing change |     132.8 | 0.39 GiB | 166.6 | 0.41 GiB |
| | After indexing change      |     83.9 | 0.42 GiB | 117.1 | 0.44 GiB |

So, around 2x faster on DAS and around 10x faster on DMAS. I've also stripped out the low-memory path now to keep this PR reviewable; we can add it back in later if we want. Better to get this part merged fast, so we can also merge the reduced precision option on top of it.

zea/tools/fit_scan_cone.py
```diff
@@ -122,8 +122,8 @@ def detect_cone_parameters(image, image_range, min_cone_half_angle_deg=20, thres
             "The image may be empty or the threshold too high."
         )
 
-    min_y = np.min(non_zero_indices[:, 0])
-    max_y = np.max(non_zero_indices[:, 0])
```
wesselvannierop, line 126: Harmless but unrelated to PR?

zea/data/convert/__main__.py
```diff
@@ -52,7 +52,7 @@ def main():
     """Parse command-line arguments and dispatch to the selected dataset conversion routine."""
     # ty cannot match tyro.cli's `TypeForm[OutT]` overload against a subcommand
     # `Union[Annotated[...], ...]`; this is tyro's documented pattern.
-    args = tyro.cli(Dataset)  # ty: ignore[no-matching-overload]
```
wesselvannierop, line 55: My ty doesnt like this, also unrelated to this PR?

wesselvannierop (changes requested):
Some changes that should be removed from this PR, ready for accept after that! Great improvements! For me on 128 focused tx carotid file, the speedup is 155 ms -> 125 ms (using `timed=True`), when using `jit_options="ops"`. Did not check the memory usage.

zea/tools/fit_scan_cone.py
Louisvh: `ty` wouldn't let me commit without it, so I just added it to be able to continue.

zea/data/convert/__main__.py
Louisvh: My `ty` gave a warning about a ty ignore statement that would not hit a ty warning or error, but this sounds like a ty version mismatch on my end.

Louisvh: Yeah, after updating my local `ty`, the warning flipped the other direction. I've reverted both those changes.

zea/tools/fit_scan_cone.py
Louisvh: updated my local environment and ty stopped complaining about this one - I've removed it in the main merge commit

tristan-deep (dismissed):
Cool analysis! And good call on splitting the PR in two. 
Nitpicks from the rabbit about adding two small tests seem relevant as well! @Louisvh

Louisvh:
Thanks! Those two tests are added now (which auto-dismissed your review; whoops). Should be ready to go I think.

tristan-deep:
> Thanks! Those two tests are added now (which auto-dismissed your review; whoops). Should be ready to go I think.

I think @wesselvannierop also requested some changes.

wesselvannierop (approved):

Issue #376, opened 2026-05-22
Invalid parameter messages pipeline


tristan-deep:
Let's investigate the following warning messages that appear quite often. For example seen in the [pipeline notebook](https://zea.readthedocs.io/en/latest/notebooks/pipeline/zea_pipeline_example.html).

```bash
zea: DEBUG [zea.Pipeline] The following input keys are not used by the pipeline: {'center_frequency', 'zlims', 'xlims'}. Make sure this is intended. This warning will only be shown once.
```

And:

```bash
zea: DEBUG Skipping invalid parameter 'n_frames'.
```

The latter one is probably fixed in #358 with the new dataspec (after converting datasets).

wesselvannierop:
@tristan-deep closed in #494?

tristan-deep:
Yes, don't see that appearing when running the notebook anymore! I double checked because I think the docs clean some of the warnings? Although not zea warnings I think. Not really sure if that was fixed in #494, but more after we improved the parameter handling and data format.

PR #494, opened 2026-07-10
Better pipeline error & warning handling

### LINKED ISSUE #57: Pipeline error handling
There are some gotchas with working with the pipeline, and especially in compiled (JIT) mode, the error messages can be cryptic. It should not be too hard to add some assertions and improved error handling to make it more user friendly. For instance already catching faulty imports at the start for example.

tristan-deep:
Closes #57 (cryptic pipeline errors, especially under JIT) and #376 (confusing "unused parameter" messages).

Pipeline errors were hard to debug. A wrong shape or dtype gave you something like `tuple index out of range`, non-JIT runs dumped the *entire* input array into the traceback, and a simple typo in a parameter name was silently swallowed. This PR makes these failures actually being printed to console.

## What changed

All in [`zea/ops/pipeline.py`](zea/ops/pipeline.py):

- **Clearer errors.** When an operation fails, the error now names the operation, keeps the original exception chain, and appends a compact `key: shape dtype` summary of the inputs — so you can see at a glance that you passed a `(10,)` array where 3+ dims were expected. The underlying message is truncated, so a concrete array is never dumped in full.
- **No more double-wrapping.** Nested pipelines used to stack `Error in operation ...: Error in operation ...`. An error is now annotated once.
- **Typo suggestions.** A missing key close to one you provided gets a hint: *"You provided 'dat', which is unused — did you mean 'data'?"*
- **Smarter unused-key warnings (#376).** Likely typos (close to a key the pipeline uses) now get a real warning with a suggestion. Genuinely extra params (e.g. leftover `zea.Parameters` fields) stay a quiet debug line instead of an alarming "warning".

## Before / after

Wrong shape — 1-D array into `EnvelopeDetect`:

```
# Before
RuntimeError: [zea.Pipeline] Error in operation 'EnvelopeDetect': tuple index out of range

# After
PipelineError: [zea.Pipeline] Operation 'EnvelopeDetect' failed with IndexError: tuple index out of range
Inputs: {data: (10,) float32}
```

Typo'd parameter — `dynamic_rang=(-50, 0)`:

```
# Before (silent, logged at debug as a benign "unused key")
DEBUG ... input keys are not used by the pipeline: {'dynamic_rang'}. Make sure this is intended.

# After
WARNING [zea.Pipeline] Some input keys look like typos and are ignored: 'dynamic_rang' -> did you mean 'dynamic_range'?
```

You can reproduce the full set of cases with snippet below:

<details open>
<summary>snippet</summary>

```python
"""Reproduce pipeline error-handling scenarios (issues #57 and #376)."""

import os

import numpy as np
import traceback

import zea
from zea.ops import Pipeline, EnvelopeDetect, Normalize, LogCompress, Identity


def banner(title):
    print("\n" + "=" * 78)
    print(title)
    print("=" * 78)


def show(title, fn):
    print(f"\n--- {title} ---")
    try:
        fn()
        print("  (no error)")
    except Exception as e:
        print(f"  {type(e).__name__}: {e}")


def make_pipeline(jit_options):
    return Pipeline(
        operations=[EnvelopeDetect(), Normalize(), LogCompress()],
        with_batch_dim=False,
        jit_options=jit_options,
    )


data = np.random.randn(64, 64, 1).astype("float32")

for jit_options in ["pipeline", "ops", None]:
    banner(f"jit_options={jit_options!r}")

    # 1. Missing required key: LogCompress reads 'data', but we pass wrong key
    def scenario_missing_key():
        p = make_pipeline(jit_options)
        p(image=data)  # wrong key name, pipeline needs 'data'

    show("1. Missing required 'data' key (typo: image=)", scenario_missing_key)

    # 2. Unused / extra parameter (issue #376)
    def scenario_unused_param():
        p = make_pipeline(jit_options)
        out = p(data=data, center_frequency=5e6, zlims=(0, 0.06), foo=123)

    show("2. Extra/unused params (center_frequency, zlims, foo)", scenario_unused_param)

    # 3. Wrong dtype: pass integer data where float expected
    def scenario_wrong_dtype():
        p = make_pipeline(jit_options)
        p(data=(data * 100).astype("int32"))

    show("3. Wrong dtype (int32 data)", scenario_wrong_dtype)

    # 4. Wrong shape: 1-D data where 3-D expected by EnvelopeDetect(axis=-3)
    def scenario_wrong_shape():
        p = make_pipeline(jit_options)
        p(data=np.random.randn(10).astype("float32"))

    show("4. Wrong shape (1-D data, axis=-3 invalid)", scenario_wrong_shape)

    # 5. Wrong-type value: dynamic_range as a string
    def scenario_bad_value():
        p = make_pipeline(jit_options)
        p(data=data, dynamic_range="loud")

    show("5. Bad value type (dynamic_range='loud')", scenario_bad_value)

    # 6. Typo in a parameter name: dynamic_rang instead of dynamic_range
    def scenario_typo():
        p = make_pipeline(jit_options)
        p(data=data, dynamic_rang=(-50, 0))

    show("6. Typo param (dynamic_rang instead of dynamic_range)", scenario_typo)

```
</details>

New tests in `tests/test_ops_infra.py` cover the input summary, typo suggestion, typo warning, and the no-double-wrap guarantee.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved pipeline failures with pipeline-aware `PipelineError`/`PipelineKeyError`, including the failing operation name and a compact `key: shape dtype` input summary.
  - Added better missing-input reporting with “did you mean?” hints for likely typos.
  - Warns on likely misspelled input names (and ignores those keys); logs benign extra inputs once.
  - Prevents double-wrapping of nested pipeline errors and adds a clearer message when string-valued inputs require prior parameter preparation.

- **Tests**
  - Added pipeline error-handling coverage for wrapping, missing-key suggestions, typo warnings, and nested exception behavior.


zea/ops/pipeline.py
```diff
@@ -410,23 +485,40 @@ def __call__(
...
+                    hints = ", ".join(f"'{k}' -> did you mean '{v}'?" for k, v in typos.items())
+                    log.warning(
+                        f"[zea.Pipeline] Some input keys look like typos and are ignored: {hints}"
+                    )
+                if benign:
+                    log.debug(
+                        f"[zea.Pipeline] Ignoring input keys not used by the pipeline: "
+                        f"{sorted(benign)}. This is expected when passing extra parameters "
+                        "(e.g. from `zea.Parameters`). Shown once per pipeline."
+                    )
```
wesselvannierop, line 520: Not sure "this is expected" is the right wording. Passing a parameter to pipeline which is confirmed not being used is still nice to know, and we should instead find why this happens

tristan-deep: First had a more nuanced message, but decided in the end to keep it simple and simply report the ignored parameters.

swpenninga (dismissed):
LGTM, all examples worked

swpenninga:
Might be worth it to look at rabbit's nitpick comments though

PR #496, opened 2026-07-13
Fix hatchling sdist bloat and polish packaging metadata

tristan-deep:
## Problem

The published sdist for `zea` had grown from ~472 KB to 35.5 MB. When the project switched build backends from poetry-core to hatchling, the wheel was properly scoped to just the `zea` package, but nothing was configured for the sdist. Hatchling's default sdist behavior is to bundle every git-tracked file, so it started pulling in `docs/` (notebook images and outputs), `tests/`, `paper/`, `configs/`, and Dockerfiles.

## Changes

- Scope the sdist to `zea/`, `README.md`, `LICENSE`, and `CITATION.cff`. This brings the sdist back down to ~504 KB; the wheel itself is unchanged.
- Add Apache-2.0 license metadata so it shows up correctly on PyPI and in license-scanning tools.
- Add a `py.typed` marker so type checkers like mypy/pyright pick up zea's type hints for downstream users.
- Add Documentation, Issues, and Contributing links to the project metadata.
- Deduplicate the `dev` optional-dependency group, which used to hand-copy every package from `tests`, `docs`, `display-headless`, `models`, and `app`. It now just references those extras directly, so they can't drift out of sync.
- Regenerate `uv.lock` to match the dependency changes above.

## Testing

Built both the sdist and wheel locally with hatchling and confirmed the sizes and file contents look right. Also diffed the resolved `dev` extra metadata before and after to confirm no dependencies were accidentally dropped or added.


wesselvannierop (approved):
Nice one

tristan-deep:
Just checked and zea v0.1.3 is now 1.1MB so issue was indeed fixed in this PR. Previously 33MB.

PR #498, opened 2026-07-13
Add best practices / FAQ

### LINKED ISSUE #272: Adding a best practices and FAQ
I notice some issues that keep occurring with people that can be resolved with some best practices. They are a bit dependent on peoples setup, that is why a section in the docs addressing this could be nice.

Some examples quickly sketched out here (these should be worked out better):

- When working with a system that has GPUs, always good to run `zea.init_device()` at the top of your script.
- When installing and working with GPU, always check the compatibility of your setup (Windows / Linux / versions etc.) with the ML library you installed (JAX / TF / Torch). You can check if the ML library sees your GPUs with: `jax.devices()` or `torch.cuda_is_available()` or `tensorflow.config.list_physical_devices()`
- When going from a numpy array on CPU to a tensor (on GPU), use `keras.ops.convert_to_tensor` and `keras.ops.convert_to_numpy`
- When running OOM during beamforming consider increasing the patches in PatchedGrid. 

Feel free to add.

tristan-deep:
Working on a best practices list (see #272). Should be curated. Working on cleaning this one up a bit. If you have any other best practices (@wesselvannierop ?) please consider adding to the TODO or directly to the list. 

TODO:
- [ ] Clean up current list (minimal slop, to the point, restructure, etc.)
- [ ] ...


docs/source/best-practices.rst
```diff
@@ -0,0 +1,286 @@
...
+   subsampled_data = data[:, parameters.selected_transmits]
+
+   image = pipeline(data=subsampled_data, **inputs)["data"]
+
+Beamforming
+-----------
+
+**Out of memory during beamforming? Increase the number of patches.**
+   The beamforming grid is processed in patches by
+   :class:`~zea.ops.PatchedGrid`. If you hit an out-of-memory (OOM) error, raise
```
wesselvannierop, line 224: I have even encountered it just hanging without ever OOM'ing, and increasing the number of patches helped

wesselvannierop (commented):
Like you said, there might be some gotcha's we can tackle with better defaults.

tristan-deep:
> Like you said, there might be some gotcha's we can tackle with better defaults.

Agreed, for reference see #547

docs/source/best-practices.rst
tristan-deep: patches are now handled differently by default in #596

tristan-deep:
Since #585 backend is selected automatically, still setting the env manually in your script is advisable to pin the backend and have a reproducible script on different machines and environments.

docs/source/best-practices.rst
wesselvannierop: Indeed, this was to fix this exact comment, which I encountered more often since, and it was not really hanging, but just a reaaaally slow compile

PR #501, opened 2026-07-14
Estimate `t_peak` from waveform

tristan-deep:
The default `t_peak` (time to pulse peak) was always `1/center_frequency`, which doesn't match reality for real Verasonics waveforms (they usually ramp up over a few cycles before peaking). 

We already had `compute_time_to_peak_stack` in `func/ultrasound.py`, so now `Parameters.t_peak` uses it when `waveforms_two_way` is available, falling back to `1/f0` otherwise. It is still possible to manually set `Parameters.t_peak`.

## Changes

- `Parameters.t_peak`: derive from `waveforms_two_way` via `compute_time_to_peak_stack` when present
- `waveforms_one_way`/`waveforms_two_way` now default to `None` in `VALID_PARAMS`
- `t_peak` now correctly invalidates its cache when `waveforms_two_way` changes (also fixes the same latent bug in `n_waveforms`)
- Added `test_t_peak_default_and_waveform_derived` in `test_scan.py`

### DIFF
--- a/tests/test_scan.py
+++ b/tests/test_scan.py
@@ -322,6 +322,54 @@ def test_accessing_valid_but_unset_attributes():
     parameters.focus_distances
 
 
+def test_t_peak_default_and_waveform_derived():
+    """t_peak falls back to 1/f0, but is estimated from waveforms_two_way when provided."""
+    center_frequency = 5e6
+    n_tx = 3
+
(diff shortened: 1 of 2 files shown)


vincentvdschaft (approved):
Nice update. I think it all looks good.

PR #503, opened 2026-07-14
Show actual values in File.summary(), not just field descriptions

tristan-deep:
`File.summary()` (and `zea data summary`) only ever printed a dataset's shape and its spec description — never the actual value. So for scalar fields like `credit` or `probe/name`, you'd see a description telling you what the field *is*, but never what it's actually set to. You had to drop into Python and read the field yourself to find out.

Now scalars and small datasets (≤ 8 elements) print their real value. Large arrays still just print shape (plus dtype now). Units/descriptions with nothing to say (the "not set" placeholder) are hidden instead of printed as noise. Output is colorized (bold keys, cyan values, dim units/descriptions).

### Example

```
zea data summary my_scan.hdf5
```

**Before:**
```
├── metadata/
│   ├── credit (shape=())
│   │   ├── description: Credit or attribution for the dataset.
│   │   └── unit: –
│   └── subject/
│       └── type/
│           ├── /metadata/subject/type (shape=())
│           │   ├── description:
│           │   ├── unit: -
```

**After:**
```
├── metadata/
│   ├── credit: "us4us Ltd."  Credit or attribution for the dataset.
│   └── subject/
│       ├── id: "s8"  Subject ID. Needed for subject-wise splits.
│       └── type: "phantom"  Subject type, e.g. human, phantom, animal.
```

Big arrays like `raw_data` or `t0_delays` are unchanged — still just shape, now with dtype too, so nothing gets dumped to the console:

```
└── raw_data (shape=(10, 1024, 470, 512, 1), dtype=float32)  Raw channel data.
```

Looks pretty too now!

image

## Additionally

Added claude hook, which basically sets up the environment and makes sure it runs `pre-commit run --all` in isolated sessions. Useful for when you would like Claude to solve a well-defined and small issue on its own. It can spin up its own environment etc. Rather it having to figure out installation and linting instructions from the docs it can use the [claude hook](https://code.claude.com/docs/en/hooks).

### DIFF
--- a/.claude/hooks/session-start.sh
+++ b/.claude/hooks/session-start.sh
@@ -1,0 +1,30 @@
+#!/bin/bash
+# Sets up a local dev environment for Claude Code on the web so linters, pre-commit,
+# and pytest all work out of the box. Only runs in remote sessions — local
+# contributors are expected to already have their own environment set up (see
+# docs/source/contributing.rst).
+set -euo pipefail
+
+if [ "${CLAUDE_CODE_REMOTE:-}" != "true" ]; then
+  exit 0
+fi
+
+cd "$CLAUDE_PROJECT_DIR"
+
+# Mirrors `uv sync --extra dev` from docs/source/contributing.rst. --frozen keeps
+# this from touching uv.lock. jax[cpu] is installed separately (not part of the
+# `dev` extra) as the cheapest backend to exercise `zea`, matching the JAX-only
+# jobs in CI.
+export UV_TORCH_BACKEND=cpu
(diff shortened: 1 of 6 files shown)


wesselvannierop:
@tristan-deep What is the .claude stuff about? Can you address that in the PR description a bit, just out of curiousity :)

wesselvannierop (approved):

tristan-deep:
> @tristan-deep What is the .claude stuff about? Can you address that in the PR description a bit, just out of curiousity :)

Added it!

PR #504, opened 2026-07-14
HDF5 chunks, compression and streaming

### LINKED ISSUE #453: Refactor `file_operations.py`
After the updated file spec, [`file_operations.py`](https://github.com/tue-bmd/zea/blob/main/zea/data/file_operations.py) got outdated and most probably broke in quite some places. Should be able to refactor the code in such a way it is robust to zea file spec changes and contains minimal logic.

wesselvannierop:
# HDF5 chunks, compression and streaming

**Loading a frame is now ~10x faster, files are ~40% smaller, and writes got faster too.** On a
real carotid scan, reading 3 frames off disk went from **1.4 s to 105 ms**; the same read streamed
from HuggingFace went from **3.0 s to 468 ms**; and just peeking at a cloud file's `summary()` went
from **2.3 s to 171 ms** — because it no longer downloads the entire 6 GB file to do it.

It started with noticing that loading a compressed HDF5 file was slow. The cause turned out to be
two things HDF5 does not do for you: it **cannot decompress chunks concurrently** (h5py decodes
them one at a time, under a global lock), and its **automatic chunk size** is a poor fit for how we
read channel data. Fixing both, and picking the codec on real data rather than by reputation, is
where the numbers above come from.

This PR:

- Switches the compression algorithm to a faster one (blosc + zstd + bitshuffle @ clevel 7) --> carotid dataset now half the disk space!
- Decompresses chunks in parallel in a thread pool, working around h5py's global lock
- Changes the default chunk size to suit how we actually read channel data
- Lets Blosc thread *within* a chunk on writes (`BLOSC_NTHREADS`), worth ~4x on writes
- Adds a streaming interface to `zea.File` for HuggingFace datasets
- Caches streamed chunks on disk, so a repeated read is ~10x faster and costs no egress
- Adds an optional progress bar for streamed reads (`zea.File(..., progress=True)`), but this only works on new/resaved files!
- Refactors the `zea data` API to be more robust by going through the `FileSpec` #453

For testing use: https://huggingface.co/datasets/zeahub/zea-carotid-2023/blob/v0.1.3/data/2_cross_bifur_right_0000.hdf5

## Speed-ups in common use cases

Just checking the summary of a cloud file:

```python
import zea
with zea.File("hf://...", stream=True) as f:
    print(f.summary())  # 171 ms, previously 2.3 s — it had to download the entire file!
```

Loading a couple of frames from a cloud file:

```python
import zea
with zea.File("hf://...", stream=True) as f:
    raw_data = f.data.raw_data[:3]  # streams chunks from cloud, decompresses in parallel
```

Loading a local file:

```python
import zea
with zea.File("scan.hdf5") as f:
    raw_data = f.data.raw_data[:3]  # concurrently reads and decompresses chunks from disk
```

## Results vs `main`

Real carotid scan (`n_tx=149`), 8 frames stored, 3 frames read. Both branches run the same data on
the same machine — `main` from its own checkout — and are asserted to read back identical values.

|                         | main (float32) | this PR      | main (int16) | this PR      |
|-------------------------|----------------|--------------|--------------|--------------|
| local read (cold cache) | 1.4 s          | **105 ms**   | 876 ms       | **87 ms**    |
| cloud `summary()`       | 1.9 s          | **145 ms**   | 1.6 s        | **163 ms**   |
| cloud read (3 frames)   | 2.8 s          | **468 ms**   | 1.9 s        | **432 ms**   |
| file size               | 713 MB         | **410 MB**   | 552 MB       | **362 MB**   |
| write                   | 203 MB/s       | **341 MB/s** | 151 MB/s     | **477 MB/s** |

That is **10–13x on local reads**, **~10x on cloud `summary()`**, **4–6x on cloud reads**, files
**34–42% smaller** — and writes come out *faster* than `main` too, despite compressing 1.5x harder.

<details>

<summary><b>Benchmark methodology, and why writes got faster</b></summary>

The script checks `main` out into a git worktree and drives each version in its own subprocess with
`PYTHONPATH` pointed at that tree, so it runs the **real code of both branches** rather than a
description of what `main` used to do.

|                | codec      | chunk                 | chunk size | stored | ratio | write        |
|----------------|------------|-----------------------|------------|--------|-------|--------------|
| main (float32) | lzf        | (1, 10, 272, 16, 1)   | 0.2 MB     | 713 MB | 1.86x | 203 MB/s     |
| **this PR**    | blosc-zstd | (1, 7, 2176, 128, 1)  | 7.8 MB     | 410 MB | 3.24x | **341 MB/s** |
| main (int16)   | lzf        | (1, 19, 272, 16, 1)   | 0.2 MB     | 552 MB | 1.20x | 151 MB/s     |
| **this PR**    | blosc-zstd | (1, 15, 2176, 128, 1) | 8.4 MB     | 362 MB | 1.83x | **477 MB/s** |

**Writes got faster, which was not obvious.** zstd at clevel 7 works far harder than lzf, so this
started out as a regression (74 MB/s against 184). HDF5 runs the filter one chunk at a time and
*single-threaded* — but Blosc can compress the blocks **within** a chunk in parallel, and its HDF5
filter reads the thread count from `BLOSC_NTHREADS` on every call. zea now defaults it to
`min(8, cpu_count)`, which turns 105 MB/s into 453 MB/s on int16. It is a `setdefault`, so an
explicit value still wins; see [`docs/source/environment.rst`](docs/source/environment.rst) for
when to turn it *down* (several dataloader workers writing at once will multiply with it).

Two notes on the cloud rows. `main` **cannot stream at all**, so reaching a cloud file means
downloading it whole — that is what its cloud numbers are, and the asymmetry is the point of the
feature. And "the cloud" is a local HTTP server with a fixed 20 ms latency per request rather than
real HF, because real HF is bandwidth-bound and noisy by 3-4x, which drowns the round-trip effect
being measured. It must be an *asyncio* server: a thread-per-connection one collapses under
concurrent range requests (462 MB/s for one range, 103 MB/s for 64), which would charge the
concurrent reader a 4x penalty invented by the harness — enough to make streaming look *slower*
than downloading. Real object stores scale up with concurrency, not down.

**Script:** [`benchmark_vs_main.py`](https://gist.github.com/wesselvannierop/fdb32e76466f124c9285aa271fa4d676)

```bash
python benchmark_vs_main.py --from-file carotid.hdf5 --n-frames 8 --read-frames 3
python benchmark_vs_main.py --from-file carotid.hdf5 --n-frames 8 --read-frames 3 --cast int16
```

</details>

<details>

<summary><b>How the codec and chunk size were chosen</b></summary>

### Chunk size is the biggest lever — and it is free

Compression ratio is **flat across chunk size** (181 MB stored whether a chunk is 1 transmit or a
whole frame), so chunk size costs nothing on disk and only ever affects read speed. A chunk is the
unit of *parallelism*: `chunk_reader` decodes each one in a single worker thread, so one 83 MB
whole-frame chunk has nothing to parallelise, while 149 single-transmit chunks drown the network in
round trips. int16 carotid, reading 1 frame:

| chunk               | local read | cloud requests | cloud read |
|---------------------|------------|----------------|------------|
| 1 tx (0.6 MB)       | 32 ms      | 149            | 1579 ms    |
| 4 tx (2.2 MB)       | **13 ms**  | 38             | 81 ms      |
| 8 tx (4.5 MB)       | 13 ms      | 19             | **77 ms**  |
| 16 tx (8.9 MB)      | 19 ms      | 10             | 122 ms     |
| whole frame (83 MB) | 102 ms     | 1              | 231 ms     |

The optimum is a broad plateau (~2–16 MB) and it is **the same plateau for local and cloud**, so one
default serves both. Hence `MAX_CHUNK_BYTES = 8 MiB`: one frame per chunk, split along `n_tx` until
it fits the budget.

### Codec: Blosc(zstd) at clevel 7 with bit-shuffle

Every codec below decodes **in-process and concurrently**, so this compares codecs rather than
"codecs my reader happens to support" (lzf is the exception, kept to show what falling back to h5py
costs). int16 carotid, 8 transmits per chunk:

| codec                    | ratio    | write MB/s | cold read MB/s | 1 frame   |
|--------------------------|----------|------------|----------------|-----------|
| none                     | 0.98     | 1509       | 3332           | 17 ms     |
| **blosc-zstd-7-bitshuf** | **1.83** | **125**    | **5670**       | **13 ms** |
| blosc-zstd-5 (byte-shuf) | 1.75     | 47         | 4561           | 20 ms     |
| blosc-zstd-9 (byte-shuf) | 1.82     | 6          | 4646           | 23 ms     |
| blosc-lz4-9              | 1.27     | 697        | 4866           | 14 ms     |
| lzf (`main`'s default)   | 1.19     | 150        | 228            | 251 ms    |

`zstd-7 + bit-shuffle` **strictly dominates** on int16 — better ratio, faster writes *and* faster
reads than `zstd-5 + byte-shuffle`. Bit-shuffle wins because an int16 has only two byte-planes for
byte-shuffle to separate, so there is far more structure to expose at the bit level. clevel 9 is a
trap: 6 MB/s writes for ~1% more compression.

Two codecs were implemented, measured, and **rejected**: Blosc2 and Bitshuffle-as-a-filter both
compress slightly better, but their Python bindings **hold the GIL** — Blosc2 scales 1.1x across
the decode thread pool where Blosc scales 7.2x, costing ~30x read throughput. No compression ratio
buys that back. (Their decoders are not in the PR; such files still read correctly via h5py.)

**Script:** [`benchmark_compression.py`](https://gist.github.com/wesselvannierop/25d248e61a26f9d86f402e73e869c3f4)

```bash
python benchmark_compression.py --from-file carotid.hdf5 --n-frames 4 --cast int16
python benchmark_compression.py --from-file carotid.hdf5 --n-frames 4 --only chunks
```

</details>

<details>

<summary><b>Follow-up: 14x faster writes still</b></summary>

Writes can be parallelised further, with threads in one process — the write-side mirror of
`chunk_reader`. Measured on the real carotid file (332 MB int16, 8.4 MB chunks, this PR's codec):

| approach                                                 | write         | vs today |
|----------------------------------------------------------|---------------|----------|
| HDF5 filter pipeline, single-threaded                    | 104 MB/s      | 1x       |
| **`BLOSC_NTHREADS=8`** (in this PR)                      | 453 MB/s      | 4.3x     |
| parallel compress + `write_direct_chunk` (16-32 threads) | **1491 MB/s** | **14x**  |

All three produce byte-identical files of the same size (181 MB), readable by plain h5py.

### Plan for `write_direct_chunk` (up for grabs)

Compression is the write bottleneck and it parallelises; the byte-writes stay serial under h5py's
lock, but they are cheap. So: compress each chunk in a `ThreadPoolExecutor`
(`numcodecs.blosc.compress` releases the GIL), then hand HDF5 the finished bytes with
`dset.id.write_direct_chunk(offset, buf)`, bypassing the filter pipeline. Working prototype (this
is the measured 1491 MB/s):

```python
ncblosc.set_nthreads(1)                       # the pool provides the parallelism, not blosc
grid = [math.ceil(s / c) for s, c in zip(raw.shape, CH)]
cells = [(i, j) for i in range(grid[0]) for j in range(grid[1])]

def compress(cell):
    i, j = cell
    offset = (i * CH[0], j * CH[1], 0, 0, 0)   # ELEMENT coords, on a chunk boundary
    sl = tuple(slice(o, min(o + c, s)) for o, c, s in zip(offset, CH, raw.shape))
    block = np.ascontiguousarray(raw[sl])
    if block.shape != CH:                      # edge chunk: HDF5 stores chunks FULL-size
        pad = np.zeros(CH, dtype=raw.dtype)
        pad[tuple(slice(0, n) for n in block.shape)] = block
        block = pad
    return offset, ncblosc.compress(block, b"zstd", 7, ncblosc.BITSHUFFLE)

with h5py.File(path, "w") as f:
    d = f.create_dataset("d", shape=raw.shape, dtype=raw.dtype, chunks=CH, **BLOSC)
    with ThreadPoolExecutor(max_workers=16) as ex:
        for offset, buf in ex.map(compress, cells):
            d.id.write_direct_chunk(offset, buf)
```

**The traps, all of which I hit building the prototype:**

1. **The codec params must match the dataset's declared filter pipeline exactly** — same cname,
   clevel, shuffle. HDF5 **does not verify this**. Get it wrong and you write a file that decodes
   to silent garbage, with no error at write *or* open. This is why it is a separate PR: write
   correctness becomes ours, and the failure mode is invisible.
2. **Offsets are element coordinates on a chunk boundary**, not chunk indices — `(i*1, j*15, 0, 0,
   0)`, not `(i, j, 0, 0, 0)`. Chunk indices raise `OSError: offset doesn't fall on chunks's
   boundary`, which at least fails loudly.
3. **Edge chunks are stored full-size.** A chunk hanging off the end of the array must be
   zero-padded to the full chunk shape before compressing, or the buffer is short.
4. **Incompressible chunks.** HDF5's filter stores such a chunk *raw* and records it in the
   `filter_mask`. Writing directly, either always store compressed (what the prototype does —
   `chunk_reader` handles the mask correctly on the way back in) or replicate the raw-store
   behaviour with the mask bit set.
5. **Memory is `workers x chunk_size`** — 16 x 8 MB here. Bound it by *bytes* the way
   `chunk_reader` bounds reads (`MAX_BYTES_IN_FLIGHT`), not by chunk count.

Diminishing returns past ~16 threads (32 was 1491, 64 was 1431 MB/s). Equality tests should mirror
`tests/data/test_chunk_reader.py`: write via the fast path, read back with **plain h5py** as the
oracle, across codecs x chunk layouts x edge-shaped arrays x incompressible data.

</details>

## Notes for reviewers

- **Existing files keep working.** lzf files still read (via h5py, serially); nothing needs
  migrating. Re-saving a file with `zea data resave` gets it the new codec and chunking.
- **`chunk_reader` correctness is ours now**, so it is a pure optimisation with an h5py fallback for
  anything it does not fully understand (unknown codec, contiguous dataset, exotic selection), and
  h5py is the oracle in the tests: equality across codecs x chunk layouts x selections, including
  incompressible chunks that HDF5 stores raw (`filter_mask`).
- **Edge case:** if a *single* transmit already exceeds `MAX_CHUNK_BYTES` (a very deep/wide
  acquisition), the chunk is left oversized rather than splitting `n_ax`/`n_el`, which are always
  read whole. Documented in `_resolve_chunks`.
- **Write speed:** `File.create` still goes through h5py, so HDF5 runs the filter **one chunk at a
  time** — chunks are not encoded concurrently. That matters less than it sounds, because Blosc
  threads *within* each chunk (`BLOSC_NTHREADS`, ~4x), which is what puts writes ahead of `main`
  despite compressing 1.5x harder. Encoding chunks in parallel *is* possible (`write_direct_chunk`,
  measured at **14x** — see the follow-up above), but it moves write correctness to us, so it is
  left for a follow-up. Anyone saving many files can already parallelise across them with a process
  pool.
- **Test file:** already updated one carotid file for testing: https://huggingface.co/datasets/zeahub/zea-carotid-2023/blob/v0.1.3/data/2_cross_bifur_right_0000.hdf5 (NOTE: not the _small file!)
- **Windows needs a fallback**, `os.pread` is not available on Windows so I needed to add a fallback for it. It is probably a bit slower compared to the Unix case, but still way faster than before. Since most our users are on unix based systems and machine learning in general happens on these systems this fix should be fine.

## Summary

* **New Features**
  * Added `hf://` lazy streaming on first access with per-chunk progress control and optional on-disk chunk caching.
  * Introduced a concurrent fast-path for reading chunked HDF5 data via direct compressed chunk fetching (including HTTP range reads).
  * Added `chunk_axes` support for file creation/resaving to improve partial/streamed frame workflows.

* **Bug Fixes**
  * Improved cache safety by treating partial cache files as misses; improved lazy remote failure behavior.

* **Documentation**
  * Expanded environment variable documentation and updated notebook setup guidance.

* **Tests**
  * Expanded chunk caching/reader, HF streaming, and data-format/layout regression coverage.


tristan-deep:
Amazing upgrades! Already have two comments / questions:

- pbar didn't seem to work for streaming? One could potentially stream multiple GBs without knowing and waiting. 
- Can you upstream the stream parameter to `zea.Dataset` and `zea.Dataloader`?

wesselvannierop:
> Amazing upgrades! Already have two comments / questions:
> 
> * pbar didn't seem to work for streaming? One could potentially stream multiple GBs without knowing and waiting.
> * Can you upstream the stream parameter to `zea.Dataset` and `zea.Dataloader`?

I set the progress bar to default off, but maybe is better to just turn it off inside a dataset/dataloader?
```python
zea.File("hf://...", progress=True)
```

When using a "hf://..." path, it streams by default now. So I guess a dataset/dataloader should stream too? I will have a look.

Edit: I'll try to make the defaults a bit nicer, like `Dataset(lazy=True)` is probably something we now always want with streaming.

zea/data/chunk_reader.py
```diff
@@ -0,0 +1,532 @@
...
+            out[index] = data
+            if self.cache is not None:
+                self.cache.put(offset, size, data)
+            if on_bytes is not None:
+                on_bytes(size)
+
+        async def gather() -> None:
+            await asyncio.gather(*(one(*miss) for miss in misses))
+
+        sync(self._fs.loop, gather)
```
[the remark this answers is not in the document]
wesselvannierop: chunk_reader.py 186-189 (cap gather by request count): the caller already bounds in-flight work by bytes (_batched(tasks, MAX_BYTES_IN_FLIGHT)), and maximal concurrency per batch is the intended design ("the ranges go out together — that is the win"). A count semaphore is a speculative enhancement, not a bug.

zea/data/chunk_reader.py
```diff
@@ -0,0 +1,532 @@
...
+        ]
+        if any(block is None for block in mapped):
+            continue
+        target = tuple(
+            block[0] for block, (_, keep) in zip(mapped, axes) if keep
+        )  # int axes collapse away
+        source = tuple(block[1] for block in mapped)
+        whole = all(block[2] for block in mapped)
+        info = dset.id.get_chunk_info_by_coord(starts)
+        tasks.append((info, target, source, whole))
```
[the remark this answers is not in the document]
wesselvannierop: chunk_reader.py 440-441 (int(info.byte_offset) on unallocated chunks): unreachable — zea writes datasets fully via create_dataset(data=...) so all chunks are allocated, and fetcher_for only serves zea hf:// files. A guard would be dead code.

tristan-deep:
> > Amazing upgrades! Already have two comments / questions:
> > 
> > * pbar didn't seem to work for streaming? One could potentially stream multiple GBs without knowing and waiting.
> > * Can you upstream the stream parameter to `zea.Dataset` and `zea.Dataloader`?
> 
> I set the progress bar to default off, but maybe is better to just turn it off inside a dataset/dataloader?
> 
> ```python
> zea.File("hf://...", progress=True)
> ```
> 
> When using a "hf://..." path, it streams by default now. So I guess a dataset/dataloader should stream too? I will have a look.
> 
> Edit: I'll try to make the defaults a bit nicer, like `Dataset(lazy=True)` is probably something we now always want with streaming.

That would be great! And yes, I think progbars are still nice most of the time given the sizes of these files, for sure when using `zea.File`? Maybe an option to disable would be the simplest so the user can choose and it is one by default?

tristan-deep:
Some tooling might be useful to include in our cli to update the zeahub dataset (`v0.1.3`) with the optimized chunking for streaming. See [script](https://gist.github.com/tristan-deep/d39d1ded45fb260a2d7cd1936382d8ba).

tristan-deep:
Some things:
- [x] On mac (outside our docker) had to install: `uv pip install requests aiohttp` for streaming (did gracefully fallback without streaming luckily!)
- [x] Rename `reading chunks` to `streaming data`? 
- [x] Older files don't allow for streaming with pbar. In that case we could still "print" downloading x GB of data? 

It is very fast now... 😮

@swpenninga could you test on Windows? Also curious to see if it resolves the issues you had with slow dataloaders in the past.

swpenninga:
> Some things:
> 
> * [x]  On mac (outside our docker) had to install: `uv pip install requests aiohttp` for streaming (did gracefully fallback without streaming luckily!)
> * [x]  Rename `reading chunks` to `streaming data`?
> * [x]  Older files don't allow for streaming with pbar. In that case we could still "print" downloading x GB of data?
> 
> It is very fast now... 😮
> 
> @swpenninga could you test on Windows? Also curious to see if it resolves the issues you had with slow dataloaders in the past.

When you say windows, do you mean locally on my laptop then? The previous issues I had were always on the linux-server, but i will give it a go

tristan-deep:
> > Some things:
> > 
> > * [x]   On mac (outside our docker) had to install: `uv pip install requests aiohttp` for streaming (did gracefully fallback without streaming luckily!)
> > * [x]   Rename `reading chunks` to `streaming data`?
> > * [x]   Older files don't allow for streaming with pbar. In that case we could still "print" downloading x GB of data?
> > 
> > It is very fast now... 😮
> > @swpenninga could you test on Windows? Also curious to see if it resolves the issues you had with slow dataloaders in the past.
> 
> When you say windows, do you mean locally on my laptop then? The previous issues I had were always on the linux-server, but i will give it a go

Yes so I guess my question was two fold:
- Test on Windows, just as a sanity check this PR works on Windows.
- Test your dataloading speed issues (note you need to also resave datasets with the new chunking otherwise the speed-gains won't be noticeable). This can be done on Linux or wherever you were developing previously.

docs/source/notebooks/models/left_ventricular_hypertrophy_segmentation_example.ipynb
```diff
@@ -39,7 +39,7 @@
    "outputs": [],
    "source": [
     "%%capture\n",
-    "%pip install zea"
+    "%pip install zea requests"
```
tristan-deep, line 42: do we need to add this requests import?

wesselvannierop: These notebooks have `import requests`, which is not a dep by zea and will fail on colab.

tristan-deep: Ah I see. Good catch!

swpenninga:
Okay heres something interesting/annoying:
- This PR uses os.pread inside of `zea/data/chunk_reader.py`
- apparently some functions of `os` only exist on specific operating systems
- So: `AttributeError: module 'os' has no attribute 'pread'` for Windows.

wesselvannierop:
@swpenninga added a fallback case for Windows, which is unfortunately a bit slower probably. Will also add a test for it in a minute.

swpenninga:
PR still fails on windows, but since @wesselvannierop has to fix this blindly, how about I just get it to work and make a commit?
I still cannot get a timing benchmark currently

wesselvannierop:
> PR still fails on windows, but since @wesselvannierop has to fix this blindly, how about I just get it to work and make a commit? I still cannot get a timing benchmark currently

You can use: https://gist.github.com/wesselvannierop/fdb32e76466f124c9285aa271fa4d676

swpenninga:
> > PR still fails on windows, but since @wesselvannierop has to fix this blindly, how about I just get it to work and make a commit? I still cannot get a timing benchmark currently
> 
> You can use: https://gist.github.com/wesselvannierop/fdb32e76466f124c9285aa271fa4d676

For Windows:
| Measurement | main | PR | Speedup |
|-------------|------|---------|---------|
| Write throughput | 88 MB/s | 82 MB/s | main slightly faster |
| Local read, 3 frames | 371 ms | 219 ms | 1.7x |
| Simulated cloud `summary()` | 2.2 s | 367 ms | 6.0x |
| Simulated cloud read, 3 frames | 2.3 s | 748 ms | 3.0x |

swpenninga:
Cant run pre-commit on this windows venv i setup, can someone quickly do it for me?

zea/data/chunk_reader.py
```diff
@@ -117,7 +117,7 @@ class LocalFetcher(Fetcher):
     per_chunk = True
 
     def __init__(self, path: str | os.PathLike):
-        self._fd: int | None = os.open(os.fspath(path), os.O_RDONLY)
+        self._fd: int | None = os.open(os.fspath(path), os.O_RDONLY | getattr(os, "O_BINARY", 0))
```
wesselvannierop, line 134: Nice, quickly checked: this is a no-op on unix 👍

wesselvannierop:
> Cant run pre-commit on this windows venv i setup, can someone quickly do it for me?

Seems to be all fine!

swpenninga (dismissed):

tristan-deep:
Pushed a few [changes](https://github.com/tue-bmd/zea/pull/504/changes/50a723ea7ec5e6dfa28c79ec688d4e70101c21db):
- Added the GB count to legacy lzf compression warning
- Changed that new files when loading from cach don't show progress bar (progbar is just for streaming / downloading)
- Print message of pbar: Stream data -> Stream <filename>
- Guard h5py kwargs a bit (I was using version but should be revision which silently failed). In future we can remove the inheritance from h5py.File altogether as @wesselvannierop proposed.

tristan-deep:
Test were failing due to the link checker (also in other PRs). You can cherry pick [this commit](https://github.com/tue-bmd/zea/pull/504/commits/ba509af95baeecad05e373ad7fff458c26a22f3f) to fix that.

tristan-deep (approved):
🚀

PR #506, opened 2026-07-16
Ultrasound computed tomography

tristan-deep:
Adds a reconstruction path for ultrasound computed tomography (USCT) acquisitions.

Supports ring arrays or dual-panel systems where individual point sources transmit in turn, rather than a wavefront steered from the receive aperture. zea's existing `TOFCorrection`/`Beamform` pipeline assumes the latter (per-element `t0_delays` and `focus_distances`) and can't represent an off-aperture point source, so USCT needs its own path.

`zea.ops.USCTReflectivityDAS` (new op, `zea/ops/usct.py`) computes a round-trip time-of-flight delay-and-sum reflectivity image: for every pixel it coherently sums the analytic channel signal across all transmit/receive pairs at the round-trip delay `‖p - source‖/c + ‖p - element‖/c - t0`.

Implementation inspired from [pyruct](https://github.com/berkanlafci/pyruct). 

> B. Lafci, J. Robin, X. L. Deán-Ben and D. Razansky.
   *Expediting Image Acquisition in Reflection Ultrasound Computed Tomography.*
   IEEE Transactions on Ultrasonics, Ferroelectrics, and Frequency Control,
   69(10):2837-2848, 2022.
   https://doi.org/10.1109/TUFFC.2022.3172713>

But then implemented with Keras / JAX and with these bonus options:
- **Transmission rejection**: masks out round-trip delays that land close to the direct through-transmission arrival, which otherwise dwarfs the backscatter.
- **Backscatter apodization**: weights each transmit/receive pair by the pixel-referred cosine between the source and element directions, keeping only geometries where the receiver looks back toward the illumination.
- An optional speed-of-sound map (`sos_map`/`sos_grid_x`/`sos_grid_z`) replaces the constant-`c` delay with a straight-ray integral of the local slowness, for cases with a known or estimated heterogeneous medium.
- **Coherent and incoheren compounding options** first one is good for media where you know the properties very well (simulations?) other one has reduced resolution but more robust against wrong sos or calibration errors.

### Coherent
phantom_938626709_z20_coherent

### Incoherent
phantom_938626709_z20_incoherent

### Coherent + SoS correction (with GT map)
phantom_938626709_z20_sos

## Summary

- **New Features**
  * Added USCT reflectivity DAS reconstruction (functional + `USCTReflectivityDAS` operation) for point-source transmit geometry, including optional heterogeneous straight-ray correction, transmission rejection, backscatter apodization, interpolation modes, and coherent vs incoherent compounding.
  * Added `channels_to_analytic` to convert RF/IQ channel data into analytic signals.

- **Refactor**
  * Renamed the receive-distance helper to `compute_receive_distances`.

- **Tests**
  * Added a comprehensive USCT regression test suite and updated receive-distance-related assertions.

- **Documentation**
  * Updated Sphinx exclusions for the new USCT APIs.


zea/func/usct.py
```diff
@@ -0,0 +1,330 @@
...
+            )
+            _, tx_unit = distance_and_unit(tx_pos, px)
+        else:
+            tx_dist, tx_unit = distance_and_unit(tx_pos, px)
+            tx_time = tx_dist / sound_speed
+
+        if per_tx_rx:
+            rx_here = receive_positions[i:j]  # (c, n_el, 2)
+            rt_list, ru_list, direct_list = [], [], []
+            for k in range(int(rx_here.shape[0])):
```
OisinNolan: This looks like it could become a vectorized map -- might enable speedup for future iterations

zea/ops/usct.py
```diff
@@ -0,0 +1,174 @@
+"""Reflectivity reconstruction for ultrasound computed tomography (USCT).
```
OisinNolan: Should this text appear in the docs somewhere? Can't find it as-is, but looks like it would be a nice addition!

OisinNolan (dismissed):
Have read through the code and didn't spot any issues, just left some minor comments. Happy to approve now since outputs look good, and we intend to iterate on this.

For future improvement it would be nice to include an example notebook + more docs -- will be a very cool addition to zea!

zea/func/usct.py
tristan-deep: Good point!

zea/ops/usct.py
tristan-deep: Yes, see they didn't get exposed after I changed it to a separate module. Now added info to the ops docstring which should be exposed :)

OisinNolan (approved):

PR #511, opened 2026-07-16
Add zea.inverse: recover channel data from beamformed images

sankethvedula:
**What changes were made?**

This PR adds `zea.inverse`, a new subpackage that inverts the DAS beamformer: given a post-beamformed image, it recovers the **pre-beamformed channel data** by expressing beamforming as a differentiable linear operator and solving a least-squares problem, optionally regularized with a point-scatterer prior.

- `zea/inverse/operators.py` — `DASOperator`: the DAS beamformer as a linear channel-data → image map, built as a regular `zea.Pipeline` (`PatchedGrid([TOFCorrection(fnum_window_fn), DelayAndSum()])`), so it shares zea's delay model (incl. lens correction) with everything else; grid-pixel patching (a `num_patches` knob) bounds memory, and the adjoint comes for free via autodiff. `ScattererSimulator`: a time-domain point-scatterer forward model driven by the scan's stored two-way waveforms, consistent with the beamformer's delay model.
- `zea/inverse/solvers.py` — matrix-free `cgls` and `linear_adjoint` (adjoint of a linear operator as the gradient of ⟨A(x), y⟩ via `zea.backend.autograd.AutoGrad`).
- `zea/inverse/seeding.py` — `seed_scatterers`: samples scatterer positions from the image envelope.
- `zea/inverse/inversion.py` — high-level drivers: `invert_direct` (minimum-norm pseudo-inverse) and `invert_scatterers` (scatterer prior with optional joint Adam refinement).
- **Supporting changes**: `TOFCorrection` accepts an `fnum_window_fn` (default unchanged); the frequency-domain simulator can run with measured waveforms via `get_measured_pulse_spectrum_fn` + a `pulse_spectrum_fn` override on `simulate_rf`/`Simulate`; a legacy-loader fix (float `tx_waveform_indices`, jaxus-style lens fields, files with only two-way waveforms); `zea.inverse` registered as a lazy submodule and in the docs.

**Three tutorials** in a new *Inversion* docs section (`docs/source/notebooks/inverse/`), ordered from controlled to clinical:

1. `fish_example.ipynb` — a realistic synthetic acquisition (Verasonics L11-5v preset, 4× sampling, five plane waves, a two-way pulse shaped by the probe's bandwidth, diffuse tissue speckle, 30 dB receiver noise). Ground truth is known exactly: the pseudo-inverse fits the image at 0.999 while recovering only 0.56 of the channel data; the scatterer prior recovers 0.66.
2. `cirs_example.ipynb` — a recorded scan of a simulated CIRS phantom, reproducing the das-inverse study: pseudo-inverse 0.556/1.000 (reference 0.557/1.000), scatterer prior 0.842/0.993 (reference 0.838/0.996).
3. `carotid_example.ipynb` — in-vivo carotid (149 single-element transmits): scatterer prior 0.060/0.931 (reference 0.057/0.946) — the honest ceiling: a compounded image constrains speckle statistics, not the channel micro-realization.

The recorded-scan tutorials run 40 CGLS iterations instead of the study's 70: on these ill-posed problems the iteration count acts as the regularization parameter, and 40 captures nearly all of the attainable correlation at a fraction of the runtime (CIRS: 17 s instead of ~8 min). Each tutorial renders a recorded-vs-recovered channel-data GIF, and the carotid one points to `zeahub/zea-carotid-2023` for zea-native carotid data (the Zenodo scan is kept for direct comparability with the study).

**Why were these changes made?**

Inverse beamforming enables working backwards from images to channel data — useful for studying what information beamforming discards, for generating physically consistent channel data from image-domain priors, and as a building block for optimization-based reconstruction. The scatterer-prior inversion follows the off-grid scatterer model of [van de Schaft et al., arXiv:2407.02285](https://arxiv.org/abs/2407.02285). Everything is written with `keras.ops` and runs on all Keras backends.

**How were the changes validated?**

- 20+ new unit tests (`tests/test_inverse.py`, `tests/test_simulator.py`, legacy-loader regression tests): adjoint identity, operator linearity, CGLS vs `lstsq`/`pinv`, simulator echo arrival times, simulator–beamformer consistency, measured-waveform spectra vs FFT, seeding properties, repeated-jit regression, and end-to-end inversions — green on jax, tensorflow and torch.
- `DASOperator` verified against the independent jaxus-based DAS implementation of the original study: measured-image correlation 1.000000.
- The two simulators cross-validate: driven by the same waveform they produce matching B-mode statistics.
- All three tutorials executed at full scale on H100/L40S GPUs; CI runs them with reduced parameters (the carotid inversion is skipped on CI).
- Full CI is green: lint, type check, tests (all backends), notebook tests incl. a new `inverse` group, docs build with `-W`, doctests, and linkcheck.

**How to test the changes**

```bash
KERAS_BACKEND=jax pytest tests/test_inverse.py tests/test_simulator.py --skip-unavailable-backends
```

or start with the introductory tutorial: `docs/source/notebooks/inverse/fish_example.ipynb`.

## Summary by CodeRabbit

* **New Features**
  * Added inverse beamforming (direct and scatterer-based) to recover channel data from DAS images, including seeded scatterer simulation and optional scatterer refinement.
  * Exposed through `zea.inverse`, with API autosummaries updated accordingly.
* **Bug Fixes**
  * Improved legacy scan loading for waveform index parsing and safer waveform reshaping when legacy fields are absent; lens-related legacy fields are ignored.
* **Documentation**
  * Updated inverse documentation and notebook/CI settings; added a Sphinx link-check exclusion and a small doc wording fix.
* **Tests**
  * Added legacy regression coverage, extensive end-to-end inversion/inverse-simulation tests, and notebook CI parameterization.


OisinNolan:
Hey @sankethvedula nice work! I had a read through the code and had some initial questions / comments, curious what your thoughts are!

1) I notice that for the `DASOperator` you built a manual delay and sum pipeline using zea functions like `calculate_delays` internally rather than using a `zea.Pipeline`. Was the main motivation for this so that you could use `keras.remat` inside to avoid OOM errors? I see that the beamforming in `DASOperator.forward` is defined recursively and sequentially over transmits -- I think this could be parallel given that the outputs of each `_beamform_tx` are summed together? This would then also apply to the gradients, since the gradient operation is linear. Maybe then you could chunk the calls to `_beamform_tx` to avoid OOMs rather than needing to use `remat`?
2) I like how you framed the scatterer simulator as a regularizer  ``min ||operator(simulate(positions, magnitudes)) - image||^2``, cool idea!
3) Very nice that you added two example notebooks on this -- they might even deserve their own `Inversion` section in the docs for visibility rather than being under `Pipeline`. One question I had about [this notebook](https://zea--511.org.readthedocs.build/en/511/notebooks/pipeline/inverse_beamforming_example.html) was about the artifacts we see in the fish simulation b-mode -- I noticed that the fish simulation in [our other simulation notebook](https://zea--511.org.readthedocs.build/en/511/notebooks/data/zea_simulation_example.html) has a similar probe / tx setup but noticeably fewer artifacts. Do you know why that is? Do you think that we could use the frequency-domain zea simulator to regularize the solution in the same way as the simulator you added? I notice in the docstring you wrote:
> For a frequency-domain simulator with parametric pulses see
    :func:`zea.simulator.simulate_rf`; this class instead uses the measured
    waveforms stored with the scan, which matters when inverting real
    acquisitions.

I'm thinking maybe the simplest approach would just be to enable the frequency-domain simulator to use a waveform stored in a scan either? Maybe we'd just need to extent `get_pulse_spectrum_fn`

4) In the real-data notebook I noticed there's quite a lot of code dedicated to parsing the file you download -- in case it's useful, we also have carotid files stored [here](https://huggingface.co/datasets/zeahub/zea-carotid-2023/tree/v0.1.3/data) in zea format, which the notebook can load directly from the URL.

tristan-deep:
Hi @sankethvedula, thanks for this contribution, looks really good!

Could you try to merge the interesting parts into a single example notebook? The multiple notebooks are appreciated as reference material (maybe hosted in a different repository, with zea version pinned at some point), but we try to keep the notebook examples in the zea docs curated. When looking at the difference in your notebooks, it is mostly various datasets, which you can easily show in one notebook I think. I see some broken link formatting in some of the notebooks as well. Lastly, try to follow the format and notebook layout / structure of some of the other notebooks. Besides showing this new inverse feature, the notebooks also serve the role of good zea practices. We'd like to reuse as many tools within the library to keep the notebooks minimal and highlighting the specific thing we want to show (in this case the inverse modeling!) Once you have a single notebook we can help with that as well to clean that up! :)

Good technical comments by @OisinNolan as well. We can host a sample on [zeahub](https://huggingface.co/zeahub/datasets) (see [other notebook examples](https://zea.readthedocs.io/en/latest/examples.html)). Maybe there is already an existing dataset on zeahub you'd like and otherwise we can upload one.

PR #513, opened 2026-07-16
Add weight and animal strain to subject metadata

OisinNolan:
ULMShare stores the weight and strain of each mouse subject. I thought these might be general enough to include as optional in `MetadataSpec.Subject`.

### DIFF
--- a/zea/data/spec.py
+++ b/zea/data/spec.py
@@ -37,7 +37,7 @@ UNITS = {
     "dB/m/Hz": "decibels per meter per hertz",
     "#": "count",
     "%": "percent",
-    "g": "grams",
+    "kg": "kilograms",
     "kg/m²": "kilograms per square meter",
 }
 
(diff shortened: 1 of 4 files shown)


zea/data/spec.py
```diff
@@ -1810,6 +1817,8 @@ class Subject(Spec):
         "type": {"unit": "–", "description": "Subject type, e.g. human, phantom, animal."},
         "age": {"unit": "–", "description": "Subject age in years.", "rare": True},
         "sex": {"unit": "–", "description": "Subject sex.", "rare": True},
+        "weight": {"unit": "kg", "description": "Subject weight.", "rare": True},
+        "strain": {"unit": "–", "description": "Animal strain, e.g. C57BL/6N.", "rare": True},
```
wesselvannierop: Might be me, but I wasn't aware what strain is, maybe the description can say that?

tristan-deep: Same for me, I was thinking: Animal genetic strain?

OisinNolan: Gotcha -- it is indeed genetic strain. What kind of description would you have in mind -- just adding the word 'genetic' to clarify or also something else?

we could also clarify via naming, e.g. `strain --> animal_genetic_strain`?

tristan-deep (approved):

PR #516, opened 2026-07-20
Improved support for curved probes

tristan-deep:
When storing data from the Verasonics with the `C5-2v`, some false warning are printed (based on assumption of a non-curved ULA). After, when reconstructing the data with `zea`, the receive aperture (through f-number) is incorrectly applied, again disregarding the curved nature of the probe. This PR aims to fix that. 

**Before**
AAApatient01_bmode_prev

**After**
AAApatient01_bmode_v2

## Details
The built-in receive f-number mask (`fnumber_mask`) decided whether an element "sees" a pixel by measuring the angle between the element→pixel direction and the global depth axis (`+z`). That is only right when every element faces straight down, i.e. a flat linear array. On a curved/convex array the peripheral elements are tilted outward, so measuring their acceptance cone from `+z` clips the aperture and throws away signal where the sector fans out.

This PR measures the acceptance cone from each element's own surface normal instead. The normal is derived directly from `probe_geometry`. F-number calculation for linear probes is unaffected. 

## API

`fnumber_mask` gains an optional `element_normals=None` argument; when omitted it derives them from `probe_geometry`. Passing explicit `+z` normals reproduces the old behaviour, which is handy as a before/after toggle in review. 

The derivation lives in a new leaf module `zea/beamform/geometry.py` (`compute_element_normals`) so the simulator and pfield directivity, which make the same `+z` assumption, can reuse it later without importing the beamformer.

## Not in this PR

- Simulator directivity (`simulator.py`) and pfield (`pfield.py`) still assume `+z`. They can use `compute_element_normals` in a follow-up; pfield is the bigger job since it currently rebuilds a synthetic flat array from pitch and ignores `probe_geometry` entirely. TODO: create issue. 

## Included Verasonics curved probes

Added two built-in Verasonics probes and a curved-geometry helper:

- **C5-2v**: 128-element convex array (pitch 0.508 mm, radius of curvature 49.57 mm), plus `create_curved_probe_geometry(n_el, pitch, radius)` (the existing `create_probe_geometry` only does flat linear arrays). The generator reproduces a real converted C5-2v file's element positions to 0.0 mm.
- **P4-2v**: 64-element flat phased array (pitch 0.30 mm).
- Mapped `C5-2v`/`C5-2V`/`P4-2v`/`P4-2V` in the converter's `_VERASONICS_TO_ZEA_PROBE_NAMES` so conversions stop falling back to the generic probe.

## Files
- `zea/beamform/geometry.py` — new, `compute_element_normals`
- `zea/beamform/beamformer.py` — `fnumber_mask` uses per-element normals
- `zea/probes.py` — `create_curved_probe_geometry` + `Verasonics_c5_2v` / `Verasonics_p4_2v` built-ins
- `zea/data/convert/verasonics.py` — C5-2v/P4-2v name mapping + `classify_ordered_geometry` (linear/arc/unknown)
- `tests/test_fnumber_mask.py` — normals correctness, linear no-op, convex FOV widening
- `tests/test_beamformer.py` — convex-probe end-to-end smoke test
- `tests/test_probes.py` — C5-2v / P4-2v geometry tests
- `tests/test_verasonics_geometry.py` — ordering-classifier tests

## Summary by CodeRabbit

* **New Features**
  * Added built-in support for Verasonics **C5-2v** (curved) and **P4-2v** (phased) probes.
  * Added curved-probe geometry generation and utilities to classify ordered probe layouts (linear/arc/unknown), plus improved Verasonics name mapping.
* **Bug Fixes**
  * Improved receive f-number masking for curved/convex probes by using per-element surface normals, widening accuracy of the illuminated field.
* **Documentation**
  * Clarified probe center frequency as the midpoint of the -6 dB band; refined probe metadata wording.
* **Tests**
  * Added/extended regression and unit tests for geometry ordering, element normals, and convex time-of-flight correction.

### DIFF
--- a/docs/source/_spec_ref.rst
+++ b/docs/source/_spec_ref.rst
@@ -969,7 +969,7 @@ Fields marked :bdg-secondary:`optional` may be absent; all others are
            - ``float32``
            - scalar
            - Hz
-           - Probe nominal centre frequency.
+           - Probe nominal centre frequency (midpoint of the probe's -6 dB band).
            - |badge-opt|
          * - ``probe_bandwidth_percent``
            - ``float32``
@@ -1107,7 +1107,7 @@ Fields marked :bdg-secondary:`optional` may be absent; all others are
             * - ``genetic_strain``
               - ``str``
               - scalar
-              - Genetic strain of an animal subject, e.g. C57BL/6N.
+              - Genetic strain (inbred line) of an animal subject, e.g. C57BL/6N.
               - |badge-opt|
             * - ``fat_percentage``
               - ``float32``
--- a/tests/test_beamformer.py
+++ b/tests/test_beamformer.py
@@ -680,6 +680,30 @@ def test_tof_correction_with_fnumber(probe_geometry, flatgrid):
 
 
 @backend_equality_check()
+def test_tof_correction_convex_probe_fnumber(flatgrid):
+    """A curved (convex) probe runs end-to-end with per-element-normal masking.
+
+    The auto-derived element normals must not produce NaNs/Infs and must still
+    mask some channels (f_number > 0) for a non-flat geometry.
+    """
+    n_el = 48
+    phi = np.linspace(-0.5, 0.5, n_el).astype(np.float32)
+    radius = 40e-3
+    convex_geometry = np.stack(
+        [radius * np.sin(phi), np.zeros_like(phi), radius * (np.cos(phi) - 1.0)], axis=-1
+    ).astype(np.float32)
+
+    inputs = _make_tof_inputs(convex_geometry, flatgrid, n_tx=2, n_ax=64)
+    inputs["f_number"] = 1.0
+    result = keras.ops.convert_to_numpy(tof_correction(**inputs))
+
+    assert result.shape == (2, flatgrid.shape[0], n_el, 1)
+    assert np.all(np.isfinite(result))
+    assert np.any(result == 0.0), "Expected some masked-out values with f_number > 0"
+    return result
+
+
+@backend_equality_check()
 def test_tof_correction_zero_data(probe_geometry, flatgrid):
     """Zero input data should produce zero output regardless of delays."""
     n_tx, n_ax = 2, 64
--- a/tests/test_fnumber_mask.py
+++ b/tests/test_fnumber_mask.py
@@ -1,5 +1,6 @@
 """Test for the fnumber_mask function."""
 
+import keras
 import numpy as np
 import pytest
 from keras import ops
@@ -10,7 +11,23 @@ from zea.beamform.beamformer import (
     fnum_window_fn_tukey,
     fnumber_mask,
 )
+from zea.beamform.geometry import compute_element_normals
 from zea.beamform.pixelgrid import cartesian_pixel_grid
+
+
+def _convex_probe_geometry(n_el=64, radius=40e-3, half_angle=0.5):
+    """Convex arc with the apex element at the origin facing +z.
+
+    Element i sits at arc-angle ``phi_i`` with position
+    ``(R sin phi, 0, R (cos phi - 1))`` and true outward normal
+    ``(sin phi, 0, cos phi)``.
+    """
+    phi = np.linspace(-half_angle, half_angle, n_el).astype(np.float32)
+    positions = np.stack(
+        [radius * np.sin(phi), np.zeros_like(phi), radius * (np.cos(phi) - 1.0)], axis=-1
+    ).astype(np.float32)
+    normals = np.stack([np.sin(phi), np.zeros_like(phi), np.cos(phi)], axis=-1).astype(np.float32)
+    return positions, normals
 
 
 @pytest.fixture
@@ -63,3 +80,96 @@ def test_fnumber_mask(probe_geometry, flatgrid, fnum_window_fn):
     idx_x = 32 + 15
     idx_z = 16
     assert mask_middle_element[idx_z, idx_x] > 0.0
+
+
+def test_compute_element_normals_flat_linear(probe_geometry):
+    """A flat linear array must derive exactly +z for every element."""
+    normals = ops.convert_to_numpy(compute_element_normals(probe_geometry))
+    expected = np.tile(np.array([0.0, 0.0, 1.0], dtype=np.float32), (probe_geometry.shape[0], 1))
+    assert np.array_equal(normals, expected)
+
+
+def test_compute_element_normals_single_element():
+    """A lone element has no tangent; fall back to +z."""
+    normals = ops.convert_to_numpy(compute_element_normals(np.zeros((1, 3), dtype=np.float32)))
+    assert np.array_equal(normals, np.array([[0.0, 0.0, 1.0]], dtype=np.float32))
+
+
+def test_compute_element_normals_convex_arc():
+    """Interior elements of a convex arc recover the analytic radial normal."""
+    positions, expected = _convex_probe_geometry(n_el=64, radius=40e-3, half_angle=0.5)
+    normals = ops.convert_to_numpy(compute_element_normals(positions))
+    # Central differences are exact in the interior; endpoints use a one-sided
+    # difference so allow them a looser tolerance.
+    np.testing.assert_allclose(normals[1:-1], expected[1:-1], atol=1e-4)
+    np.testing.assert_allclose(normals[[0, -1]], expected[[0, -1]], atol=2e-2)
+
+
+@pytest.mark.parametrize(
+    "fnum_window_fn", [fnum_window_fn_hann, fnum_window_fn_rect, fnum_window_fn_tukey]
+)
+def test_fnumber_mask_flat_linear_is_legacy_noop(probe_geometry, flatgrid, fnum_window_fn):
+    """On a flat array the per-element-normal mask equals the old global-+z mask.
+
+    The reference is the legacy angle-relative-to-global-+z formula computed
+    with the same backend ops, so any difference would be a real change in
+    behaviour rather than a norm-implementation rounding difference.
+    """
+    rel = flatgrid[:, None] - probe_geometry[None]
+    rel_norm = ops.linalg.norm(rel, axis=-1)
+    alpha = ops.arccos(rel[..., 2] / (rel_norm + 1e-6))
+    max_alpha = ops.arctan(1.0 / (2.0 * 0.5 + keras.backend.epsilon()))
+    reference = ops.convert_to_numpy(fnum_window_fn(alpha / max_alpha))[..., None]
+
+    mask = ops.convert_to_numpy(
+        fnumber_mask(flatgrid, probe_geometry, f_number=0.5, fnum_window_fn=fnum_window_fn)
+    )
+    assert np.array_equal(mask, reference)
+
+
+def test_fnumber_mask_convex_widens_field_of_view():
+    """Per-element normals extend a convex array's field of view vs forcing +z.
+
+    Because peripheral elements of a convex array physically look outward,
+    steering each element's acceptance cone along its true normal illuminates
+    sector-edge pixels that the global-+z cone blacks out. That shows up as a
+    larger field of view (more pixels receiving any aperture), which is the
+    "larger frame" a curved probe should produce.
+    """
+    # Odd element count so a real element sits exactly at arc-angle 0 (normal +z).
+    positions, _ = _convex_probe_geometry(n_el=65, radius=40e-3, half_angle=0.5)
+    z_normals = np.tile(np.array([0.0, 0.0, 1.0], dtype=np.float32), (positions.shape[0], 1))
+
+    # A grid spanning the full sector so the tilted elements' look-direction
+    # pixels are actually present.
+    grid = (
+        cartesian_pixel_grid(
+            xlims=(-30e-3, 30e-3), zlims=(1e-3, 55e-3), grid_size_x=160, grid_size_z=160
+        )
+        .reshape(-1, 3)
+        .astype(np.float32)
+    )
+
+    mask_new = ops.convert_to_numpy(
+        fnumber_mask(grid, positions, f_number=2.0, fnum_window_fn=fnum_window_fn_rect)
+    )[..., 0]
+    mask_old = ops.convert_to_numpy(
+        fnumber_mask(
+            grid,
+            positions,
+            f_number=2.0,
+            fnum_window_fn=fnum_window_fn_rect,
+            element_normals=z_normals,
+        )
+    )[..., 0]
+
+    # Per-pixel aperture count -> field of view is the set of illuminated pixels.
+    aperture_new = mask_new.sum(axis=1)
+    aperture_old = mask_old.sum(axis=1)
+    assert int((aperture_new > 0).sum()) > int((aperture_old > 0).sum())
+    # Some sector-edge pixels are illuminated only once the normals are correct.
+    assert np.any((aperture_new > 0) & (aperture_old == 0))
+
+    # The apex element (arc-angle 0) faces +z, so its column is unchanged.
+    apex = positions.shape[0] // 2
+    np.testing.assert_allclose(mask_new[:, apex], mask_old[:, apex], atol=1e-6)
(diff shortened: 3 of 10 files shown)


tristan-deep:
@coderabbitai create an issue, see the PR description, to fix element directivity in simulator / pfield to take into account the element normal as we do during f-number calculation in this PR (out of scope to also fix in this PR).

swpenninga:
image

I dont know if the whole implementation is correct now, but I could verify with a simulated phantom. Sidelobes seem pretty large at the edges

tristan-deep:
@coderabbitai review

tristan-deep:
Based on @swpenninga [idea](https://github.com/tue-bmd/zea/pull/516#issuecomment-5020041744), generated bit more insights with point scatters. See snippet to reproduce below as well. 

curved_probe_fnumber_review


<details closed>
<summary>Snippet</summary>

```python
"""Review snippet for the curved-probe f-number fix (PR "f-number curved array", eaf6f1a7).

Simulates point-scatterer RF data with the Verasonics C5-2v convex array and
reconstructs it with two beamforming pipelines that are IDENTICAL except for
one thing: how the receive f-number cone is measured for a curved array.

- "before": the f-number cone is measured against the global +z axis for every
  element, which is what zea did before this PR (correct for a flat linear
  array, wrong for a convex one since peripheral elements physically tilt
  outward).
- "after": the cone is measured against each element's own surface normal
  (:func:`zea.beamform.geometry.compute_element_normals`), which is the fix.

Both pipelines run Simulate -> Demodulate -> Beamform -> EnvelopeDetect ->
Normalize -> LogCompress end to end, at f_number=1.5. Point scatterers are
placed across the C5-2v's angular sector, including near the edges, since
that's exactly where the pre-fix cone clips the receive aperture (see
``tests/test_fnumber_mask.py::test_fnumber_mask_convex_widens_field_of_view``).
"""

import os

from contextlib import contextmanager, nullcontext

import numpy as np
from keras import ops
from mpl_toolkits.axes_grid1 import make_axes_locatable

import zea.beamform.beamformer as beamformer_module
from zea import Parameters, Probe, init_device
from zea.beamform.beamformer import fnum_window_fn_rect, fnumber_mask
from zea.beamform.geometry import compute_element_normals
from zea.display import to_8bit
from zea.ops import (
    ApplyWindow,
    Beamform,
    Cast,
    Demodulate,
    EnvelopeDetect,
    LogCompress,
    Normalize,
    Pipeline,
    Simulate,
)
from zea.visualize import set_mpl_style

init_device(verbose=False)
set_mpl_style()

F_NUMBER = 1.5
SOUND_SPEED = 1540.0
# Tighter than the usual -50 dB: this is a sparse point-target scene with no
# speckle background, so grating-lobe energy that a real speckle image would
# hide shows up much more readily. See the printed note below for why those
# grating lobes are there in the first place (it's physical, not a bug).
DYNAMIC_RANGE = (-40, 0)

# ---------------------------------------------------------------------------
# Monkeypatch reproducing the pre-fix behaviour: force every element's normal
# to global +z, exactly like `fnumber_mask` did before `compute_element_normals`
# existed. Restored afterwards so "after" runs the real, current code path.
# ---------------------------------------------------------------------------


@contextmanager
def legacy_global_z_fnumber():
    """Patch `compute_element_normals` to the pre-fix global-+z-for-everyone behaviour."""
    original = beamformer_module.compute_element_normals

    def _global_z_normals(probe_geometry, eps=1e-12):
        z_hat = ops.cast(ops.convert_to_tensor([0.0, 0.0, 1.0]), probe_geometry.dtype)
        return z_hat[None] * ops.ones_like(probe_geometry)

    beamformer_module.compute_element_normals = _global_z_normals
    try:
        yield
    finally:
        beamformer_module.compute_element_normals = original


# ---------------------------------------------------------------------------
# Probe + transmit scheme: C5-2v convex array, single simultaneous-fire
# transmit. Every element fires at t=0; on a flat array that's a plane wave,
# but the arc geometry of a convex array makes it diverge on its own, which is
# the classic single-transmit convex-probe acquisition.
# ---------------------------------------------------------------------------
probe = Probe.from_name("verasonics_c5_2v")
n_el = probe.n_el

# `Probe.get_pitch` only supports flat linear geometries; for the curved C5-2v
# the pitch is just the (constant) arc distance between neighbouring elements.
pitch = float(np.linalg.norm(probe.probe_geometry[1] - probe.probe_geometry[0]))
wavelength = SOUND_SPEED / probe.probe_center_frequency
print(
    f"Element pitch = {pitch * 1e3:.3f} mm vs. wavelength/2 = {wavelength / 2 * 1e3:.3f} mm "
    f"at {probe.probe_center_frequency / 1e6:.1f} MHz -- pitch > lambda/2, so this convex array "
    "is spatially undersampled and will show grating lobes (a real property of the C5-2v, not "
    "a simulation bug). A single uncompounded, unfocused full-aperture transmit onto a sparse "
    "point-target scene (no speckle to mask them in) makes this especially visible; real "
    "acquisitions hide it behind compounding, tissue speckle, and elevation/lens focusing."
)

t0_delays = np.zeros((1, n_el), dtype=np.float32)
# Hann-tapered transmit aperture: reduces edge-diffraction clutter from the
# unfocused full-aperture transmit so the reconstructions stay readable; it
# does not affect the receive f-number comparison, which is the point here.
tx_apodizations = np.hanning(n_el)[None].astype(np.float32)
polar_angles = np.zeros(1, dtype=np.float32)
focus_distances = np.full(1, np.inf, dtype=np.float32)


def aperture_element_count(points, legacy):
    """Number of elements contributing to each point under the old (legacy=True,
    global +z) or new (legacy=False, per-element normal) f-number cone."""
    ctx = legacy_global_z_fnumber() if legacy else nullcontext()
    with ctx:
        mask = ops.convert_to_numpy(
            fnumber_mask(
                points.astype(np.float32), probe.probe_geometry, F_NUMBER, fnum_window_fn_rect
            )
        )[..., 0]
    return (mask > 0).sum(axis=1)


# ---------------------------------------------------------------------------
# Point-scatterer phantom: an angle x depth grid spanning the C5-2v's angular
# sector (half-angle ~37 deg for this probe geometry, derived below from the
# tilt of its outermost element), plus one scatterer specifically placed where
# the pre-fix +z cone gave it *zero* receive aperture (fully outside the old
# FOV) but the fix gives it a usable aperture -- found by scanning candidate
# angle/depth points with the actual before/after f-number mask rather than
# assuming a specific angle up front.
# ---------------------------------------------------------------------------
element_normals = ops.convert_to_numpy(compute_element_normals(probe.probe_geometry))
half_angle_rad = float(np.arctan2(np.abs(element_normals[-1, 0]), element_normals[-1, 2]))
print(f"C5-2v angular sector half-angle (from element tilt): {np.degrees(half_angle_rad):.1f} deg")

grid_angles_deg = np.linspace(-0.85, 0.85, 7) * np.degrees(half_angle_rad)
grid_depths = np.array([25e-3, 45e-3, 65e-3, 85e-3])
angle_mesh, depth_mesh = np.meshgrid(np.deg2rad(grid_angles_deg), grid_depths)
grid_x = (depth_mesh * np.sin(angle_mesh)).ravel()
grid_z = (depth_mesh * np.cos(angle_mesh)).ravel()

# Search for the "outside old FOV" scatterer. The old bug is about each
# element's *own* look direction, so search along rays from the outermost
# elements (rather than sweeping angle/depth from the apex, which never lands
# exactly on zero old-aperture: the old vertical cone from *neighbouring*
# elements still reaches most points near the sector edge). Keep candidates
# within the displayed grid and depth range so the annotated point is nicely
# on-screen.
edge_element_idx = np.arange(n_el - 45, n_el)
search_ranges = np.linspace(5e-3, 70e-3, 60)
ei, ri = np.meshgrid(edge_element_idx, search_ranges, indexing="ij")
candidates = (
    probe.probe_geometry[ei.ravel()] + ri.ravel()[:, None] * element_normals[ei.ravel()]
).astype(np.float32)
old_counts = aperture_element_count(candidates, legacy=True)
new_counts = aperture_element_count(candidates, legacy=False)
# Also keep it a few mm away (in depth) from the grid rows above, since a
# depth-matched strong grid scatterer's grating-lobe streaks would otherwise
# cross right through it and swamp its (much weaker, ~1/4-aperture) echo.
min_depth_gap = np.min(np.abs(candidates[:, 2:3] - grid_depths[None, :]), axis=1)
recovered = (
    (old_counts == 0)
    & (new_counts >= 15)
    & (candidates[:, 2] > 25e-3)
    & (candidates[:, 2] < 75e-3)
    & (np.abs(candidates[:, 0]) < 48e-3)
    & (min_depth_gap > 6e-3)
)
if not recovered.any():
    raise RuntimeError("No candidate point found that is invisible before and visible after.")
recovered_idx = np.argmax(np.where(recovered, new_counts, -1))
recovered_point = candidates[recovered_idx]
print(
    f"Edge-case scatterer at x={recovered_point[0] * 1e3:.1f} mm, z={recovered_point[2] * 1e3:.1f} mm: "
    f"{old_counts[recovered_idx]} elements before fix, {new_counts[recovered_idx]} after."
)

# Boosted reflectivity: this point only ever gets a partial receive aperture
# (~1/4 of the elements, even after the fix, since it sits right at the edge
# of the sector), so on an equal footing with the fully-illuminated grid
# points its echo is easily lost under other targets' grating-lobe clutter.
RECOVERED_MAGNITUDE = 12.0
scatterer_positions = np.concatenate(
    [
        np.stack([grid_x, np.zeros_like(grid_x), grid_z], axis=1),
        recovered_point[None],
    ],
    axis=0,
).astype(np.float32)
scatterer_magnitudes = np.concatenate(
    [np.ones(len(grid_x), dtype=np.float32), np.array([RECOVERED_MAGNITUDE], dtype=np.float32)]
)

# ---------------------------------------------------------------------------
# Scan parameters: wide Cartesian grid covering the full sector, so the FOV
# difference is visible directly without needing scan conversion.
# ---------------------------------------------------------------------------
xlims = (-55e-3, 55e-3)
zlims = (1e-3, 90e-3)

parameters = Parameters(
    n_tx=1,
    n_el=n_el,
    center_frequency=probe.probe_center_frequency,
    sampling_frequency=probe.probe_center_frequency * 4,
    t0_delays=t0_delays,
    tx_apodizations=tx_apodizations,
    element_width=probe.element_width,
    focus_distances=focus_distances,
    polar_angles=polar_angles,
    initial_times=np.ones(1, dtype=np.float32) * 1e-6,
    n_ax=2048,
    xlims=xlims,
    zlims=zlims,
    grid_size_x=500,
    grid_size_z=405,
    n_ch=1,
    selected_transmits="all",
    sound_speed=SOUND_SPEED,
    apply_lens_correction=False,
    lens_thickness=1e-3,
    lens_sound_speed=1000.0,
    # Realistic soft-tissue attenuation. With a sparse point-target scene (no
    # speckle) and a single unfocused/uncompounded transmit, off-axis grating-
    # lobe energy (see note below) is otherwise very visible; attenuation rolls
    # it off with depth like it would in a real acquisition.
    attenuation_coef=0.5,
    f_number=F_NUMBER,
)


def make_pipeline():
    """Simulate -> beamform pipeline, all in one.

    Each call builds fresh operation instances, so its JIT cache is private to
    this pipeline object: tracing under `legacy_global_z_fnumber()` bakes in
    the patched behaviour for *this* compiled function only, and a pipeline
    built outside the patch traces (and caches) the real, current code path.
    """
    return Pipeline(
        operations=[
            Simulate(),
            Cast(dtype="float32"),
            ApplyWindow(),
            Demodulate(),
            Beamform(beamformer="delay_and_sum", num_patches=20),
            EnvelopeDetect(),
            Normalize(),
            LogCompress(),
        ],
        with_batch_dim=False,
        jit_options="pipeline",
    )


def run_pipeline():
    pipeline = make_pipeline()
    inputs = pipeline.prepare_parameters(
        parameters,
        scatterer_positions=scatterer_positions,
        scatterer_magnitudes=scatterer_magnitudes,
    )
    outputs = pipeline(**inputs, dynamic_range=DYNAMIC_RANGE)
    return outputs[pipeline.output_key]


print("Running 'before' pipeline (f-number cone measured against global +z)...")
with legacy_global_z_fnumber():
    image_before = run_pipeline()

print("Running 'after' pipeline (f-number cone measured against each element's own normal)...")
image_after = run_pipeline()

# ---------------------------------------------------------------------------
# Receive-aperture coverage maps: same comparison, but directly on the
# f-number mask rather than through the full simulate/beamform pipeline. Shows
# the field-of-view widening the reconstructions above are a consequence of.
# ---------------------------------------------------------------------------
flatgrid = parameters.flatgrid
grid_shape = parameters.grid.shape[:2]  # (grid_size_z, grid_size_x)

mask_after = ops.convert_to_numpy(
    fnumber_mask(flatgrid, probe.probe_geometry, F_NUMBER, fnum_window_fn_rect)
)[..., 0]
with legacy_global_z_fnumber():
    mask_before = ops.convert_to_numpy(
        fnumber_mask(flatgrid, probe.probe_geometry, F_NUMBER, fnum_window_fn_rect)
    )[..., 0]

aperture_before = (mask_before > 0).sum(axis=1).reshape(grid_shape)
aperture_after = (mask_after > 0).sum(axis=1).reshape(grid_shape)

n_fov_before = int((aperture_before > 0).sum())
n_fov_after = int((aperture_after > 0).sum())
print(
    f"Illuminated pixels (aperture > 0): before={n_fov_before}, after={n_fov_after} "
    f"(+{n_fov_after - n_fov_before}, {100 * (n_fov_after - n_fov_before) / n_fov_before:.1f}%)"
)

# ---------------------------------------------------------------------------
# Figure: B-mode reconstructions on top, receive-aperture coverage maps below.
# ---------------------------------------------------------------------------
extent = parameters.extent_imshow * 1e3  # m -> mm
probe_x_mm = probe.probe_geometry[:, 0] * 1e3
probe_z_mm = probe.probe_geometry[:, 2] * 1e3
recovered_x_mm, recovered_z_mm = recovered_point[0] * 1e3, recovered_point[2] * 1e3

fig, axes = plt.subplots(3, 2, figsize=(10, 16), sharex="row", sharey="row")

for ax, image, title in zip(
    axes[0],
    [image_before, image_after],
    ["Before fix (cone vs. global +z)", "After fix (cone vs. element normal)"],
):
    ax.imshow(to_8bit(image, dynamic_range=DYNAMIC_RANGE), cmap="gray", extent=extent)
    ax.plot(probe_x_mm, probe_z_mm, "y.", markersize=2, label="probe elements")
    rect = plt.Rectangle(
        (recovered_x_mm - 15, recovered_z_mm - 15), 30, 30, ec="tab:red", fc="none", lw=1
    )
    ax.add_patch(rect)
    ax.set_title(title)
    ax.set_xlabel("X (mm)")
axes[0, 0].set_ylabel("Z (mm)")
axes[0, 0].legend(loc="lower left", fontsize=7)

# Zoomed-in crop of the same reconstructions around the "outside old FOV"
# scatterer (red box above), so it is actually visible rather than just
# guessed at from an arrow.
for ax, image, edge_label in zip(
    axes[1],
    [image_before, image_after],
    ["invisible before fix (0 elements)", "recovered after fix (edge of FOV)"],
):
    ax.imshow(to_8bit(image, dynamic_range=DYNAMIC_RANGE), cmap="gray", extent=extent)
    ax.plot(recovered_x_mm, recovered_z_mm, "s", mfc="none", mec="tab:red", markersize=20, mew=1)
    ax.set_xlim(recovered_x_mm - 15, recovered_x_mm + 15)
    ax.set_ylim(recovered_z_mm + 15, recovered_z_mm - 15)
    ax.set_title(edge_label)
    ax.set_xlabel("X (mm)")
axes[1, 0].set_ylabel("Z (mm)")

aperture_vmax = max(aperture_before.max(), aperture_after.max())
for ax, aperture, title in zip(
    axes[2],
    [aperture_before, aperture_after],
    ["Receive aperture count (before)", "Receive aperture count (after)"],
):
    im = ax.imshow(aperture, cmap="viridis", extent=extent, vmin=0, vmax=aperture_vmax)
    ax.set_title(title)
    ax.set_xlabel("X (mm)")
axes[2, 0].set_ylabel("Z (mm)")

# Attach the colorbar to its own axis, sized to exactly match the subplot
# height, rather than a `fig.colorbar(..., shrink=...)` guess.
divider = make_axes_locatable(axes[2, 1])
cax = divider.append_axes("right", size="5%", pad=0.15)
fig.colorbar(im, cax=cax, label="# elements contributing")

fig.suptitle(f"C5-2v convex array, f-number={F_NUMBER} -- curved-array f-number fix")
fig.tight_layout(rect=[0, 0, 1, 0.97])

out_path = "curved_probe_fnumber_review.png"
fig.savefig(out_path, dpi=150, bbox_inches="tight")
print(f"Saved comparison figure to {out_path}")

```

</details>

PR #538, opened 2026-07-31
Verasonics multibuffer

lvknippenberg:
# [Feature]: Verasonics converter — multi-buffer RF + external `.bin` RF support

Wessel mentioned that my fast saving method (dumping RF data in .bin files) and support for multiple RF buffers could be valuable additions to the toolbox. Please let me know what you think and what changes would be required. 

## What changed?

Extends the Verasonics converter (`zea/data/convert/verasonics.py`) to convert
acquisitions with **multiple RF buffers** and to read RF from **external binary
files**, while keeping all existing single-buffer behaviour intact. Adds
standalone tests and documentation.

- **Multiple RF buffers.** A single `.mat` whose `RcvData` holds more than one
  cell is now convertible one buffer at a time. `to_zea(buffer_index=...)`
  selects a buffer; `to_zea_all()` converts every buffer from one open file
  (the shared `Receive`/`Event` reads are memoised, so the cost is paid once).
  The RF **image buffer** (`ImgDataP` / `Resource.ImageBuffer`) is indexed
  independently via `image_buffer_index`.
- **External `.bin` RF.** For large acquisitions the RF of each buffer can be
  saved to a raw `int16` file `RF_data_{k}.bin` next to a metadata `.mat` that
  stores the per-buffer dimensions (`RF_rows`, `RF_cols`, `RF_frames`) and the
  saved-channel masks (`NonzeroRFcolumns`). The converter reconstructs the
  full-channel array from these, so everything downstream is identical to the
  in-`.mat` path.
- **CLI dispatch.** A source directory containing `RF_data_*.bin` is detected as
  one multi-buffer acquisition and converted per buffer; a directory of plain
  `.mat` files keeps the existing one-file-per-`.mat` behaviour. The metadata
  `.mat` may have any name (exactly one must be present).
- **Multi-TGC files.** Acquisitions that store one TGC waveform per buffer
  (`TGC.Waveform` as an array of references) are handled by selecting the
  waveform each buffer's receives use (`Receive.TGC`). Single-TGC files
  (plain matrix) are unchanged.
- **Robust reference detection.** Cell / struct-array datasets are detected via
  `h5py.check_dtype(ref=...)` instead of inspecting `fillvalue`, which is only a
  null reference for writers that set it explicitly (MATLAB does). No behaviour
  change on real MATLAB files; also makes the read paths unit-testable with
  synthetic fixtures.
- **Robust channel selection.** `Trans.ConnectorES` is applied whenever the
  acquired channel count differs from the element count (previously only for the
  updated `save_raw` format), fixing an `n_el` mismatch that also affected some
  single-buffer files.
- **Windows logging fix (`zea/log.py`).** The console stream is reconfigured
  with `errors="backslashreplace"` so Unicode log characters (arrows,
  checkmarks) don't raise `UnicodeEncodeError` on legacy-codepage (cp1252)
  consoles.
- **Tests (`tests/test_verasonics_multibuffer.py`).** New standalone unit tests
  (no external data) that build tiny synthetic MATLAB-v7.3 HDF5 workspaces in
  `tmp_path` and verify per-buffer RF reading for both storage forms.
- **Docs (`docs/source/data-acquisition.rst`).** The Verasonics section now
  documents the multi-buffer API and the external `.bin` workflow, including the
  on-disk format contract and a vendored reference MATLAB saver
  (`docs/source/_examples/WriteRFbuffersBin.m`) shown via `literalinclude`.

## Why?

Cardiac ARF / shear-wave acquisitions produce several RF buffers per recording
(e.g. a wide-beam B-mode buffer plus an active shear-wave tracking buffer, or
subsequent measurements such as imaging then Doppler), and writing the full
`RcvData` to a v7.3 `.mat` is prohibitively slow for large buffers — hence the
external `.bin` workflow. The previous converter could only handle a single
in-`.mat` buffer, and tripped over per-buffer TGC waveforms and an 80-vs-128
channel mismatch on real probes.

## How it addresses the issue

Buffer selection is threaded through the read path (`read_transmit_events`,
`read_scan`, `read_raw_data`, sampling/demod/TGC accessors) via a 0-based
`buffer_index`, resolved from `Receive.bufnum`. `read_raw_buffer_array` is the
single entry point that returns one buffer as `(n_frames, n_channels, n_samples)`
from either `RcvData` or `RF_data_{k}.bin`, so the rest of the pipeline is
storage-agnostic.

## Files

- `zea/data/convert/verasonics.py` — converter changes.
- `zea/log.py` — Windows Unicode logging fix.
- `tests/test_verasonics_multibuffer.py` — new standalone unit tests.
- `docs/source/data-acquisition.rst` — Verasonics multi-buffer / `.bin` docs.
- `docs/source/_examples/WriteRFbuffersBin.m` — vendored reference MATLAB saver.
  *(Placement is provisional — happy to move it to an `examples/` tree or drop
  it if maintainers prefer; see "Open questions" below.)*

## Testing

Standalone unit tests in `tests/test_verasonics_multibuffer.py`:
- multi-cell `RcvData`: buffer enumeration, `Receive.bufnum` → buffer mapping,
  correct per-buffer read;
- external `.bin`: full-channel reconstruction via `NonzeroRFcolumns`,
  "defined but not saved" buffers, dimension-mismatch error.

Run:

```bash
pip install -e .[dev]
pre-commit run --all-files
pytest -m 'not heavy and not notebook' --skip-unavailable-backends
# targeted:
pytest tests/test_verasonics_multibuffer.py tests/test_verasonics_geometry.py \
       tests/test_log.py tests/data/test_convert_main.py --skip-unavailable-backends
# real-file regression (heavy):
pytest "tests/data/test_conversion_scripts.py::test_conversion_script[verasonics]"
```

Locally: `ruff check`, `ruff format --check`, and `ty check` are clean on the
changed files; the targeted tests and the heavy
`test_conversion_script[verasonics]` pass (confirming the robust reference
detection does not change behaviour on real MATLAB files). The docs section was
rendered with an isolated Sphinx build (`sphinx_design`, `-W --keep-going`): no
warnings; the dropdowns, `matlab` code blocks and the `literalinclude` all
render.

Also verified end-to-end on a real 2-buffer acquisition saved both ways
(external `.bin` vs conventional `.mat`): the converted RF matches between the
two (identical shapes, dtypes, sampling frequency and `n_tx`; near-identical
values), including the baseband-IQ buffer.

## Open questions for maintainers

- **Where should the reference MATLAB saver live?** It is currently under
  `docs/source/_examples/` and pulled into the docs with `literalinclude`. An
  `examples/verasonics/` tree (or omitting the file and keeping only the format
  contract in the docs) are equally fine — happy to move it.

## Summary by CodeRabbit

- **New Features**
  - Added support for converting Verasonics acquisitions containing multiple RF buffers.
  - Added support for externally stored per-buffer binary RF data, including channel-mask restoration and zero-filled channels.
  - Added options to select RF and image buffers independently.
  - Added conversion of individual buffers or all buffers with configurable overwrite and error handling.

- **Documentation**
  - Added guidance for multi-buffer conversion, external binary layouts, metadata requirements, and MATLAB acquisition callbacks.

### DIFF
--- a/docs/source/_examples/WriteRFbuffersBin.m
+++ b/docs/source/_examples/WriteRFbuffersBin.m
@@ -1,0 +1,174 @@
+%WriteRFbuffersBin  Save Verasonics RF buffers as raw .bin files + one .mat.
+%
+% Replaces the slow '-v7.3' save of the full RcvData: each RF buffer is written
+% to a raw int16 binary file, and only the (small) dimensions and metadata
+% needed to reconstruct it are stored in a single Parameters.mat next to the
+% binaries.
+%
+% OUTPUT FORMAT (read by zea's Verasonics converter,
+% zea/data/convert/verasonics.py):
+%   <RF_data_datetime>/
+%     Parameters.mat            all base-workspace variables (Trans, TX, TW,
+%                               Receive, Event, SeqControl, Resource, TGC, ...)
+%                               EXCEPT the bulky RcvData/IQ/image buffers, plus:
+%       RF_rows   (1,nBuf)      fast-time samples per buffer (RcvData rows)
+%       RF_cols   (1,nBuf)      number of SAVED channels per buffer
+%       RF_frames (1,nBuf)      frames per buffer
+%       NonzeroRFcolumns {1,nBuf}  per-buffer logical channel mask; its length
+%                               is the full hardware-channel count and it is
+%                               true for every channel written to the binary.
+%       Resource, ImgDataP      re-added (the converter reads frame order from
+%                               Resource and the reference image from ImgDataP).
+%     RF_data_<k>.bin           buffer k (1-based, matching RcvData{k}) as int16
+%                               in MATLAB column-major order
+%                               (rows x savedChannels x frames).
+%   A buffer that is entirely zero (allocated but never acquired) is NOT written,
+%   so the converter treats it as "present but not saved" rather than reading a
+%   zero buffer.
+%
+% When to call:
+%   1. After the acquisition and after VSX exits - all buffers are then in the
+%      base workspace.
+%   2. From a UI pushbutton during runtime.
+%   3. From an external-function call in the event loop.
+%   In cases 2/3 the buffers are copied into this function's workspace first via
+%   runAcq('copyBuffers'). We detect case 1 by checking whether RcvData already
+%   exists in the base workspace.
+%
+% Wiring a save button in a Verasonics setup script:
+%   import vsv.seq.uicontrol.VsButtonControl
+%   UI(n).Control  = VsButtonControl('LocationCode','UserB6','Label','Save RF');
+%   UI(n).Callback = @WriteRFbuffersBin;
+
+function [] = WriteRFbuffersBin(varargin)
+% varargin absorbs the UI action a Verasonics pushbutton passes to its callback.
+
+SaveIQ = false;   % also store IData/QData (larger files); RF is always saved.
+
+tic;
+
+% --- Obtain the buffers -------------------------------------------------------
+% After VSX exits the buffers are in the base workspace; during a live
+% acquisition (UI button / external function) they must be copied here first.
+% ImgData/ImgDataP/IData/QData are optional: not every sequence produces them.
+if evalin('base', 'exist(''RcvData'',''var'')')
+    RcvData  = evalin('base', 'RcvData');
+    Resource = evalin('base', 'Resource');
+    [ImgData, hasImgData] = getBaseVar('ImgData');
+    [ImgDataP, hasImgDataP] = getBaseVar('ImgDataP');
+    if SaveIQ
+        [IData, hasIData] = getBaseVar('IData');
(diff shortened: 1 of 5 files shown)


docs/source/_examples/WriteRFbuffersBin.m
```diff
@@ -0,0 +1,174 @@
+%WriteRFbuffersBin  Save Verasonics RF buffers as raw .bin files + one .mat.
+%
+% Replaces the slow '-v7.3' save of the full RcvData: each RF buffer is written
+% to a raw int16 binary file, and only the (small) dimensions and metadata
+% needed to reconstruct it are stored in a single Parameters.mat next to the
+% binaries.
```
wesselvannierop, line 6: Forgot to mention we keep zea python-only, so while the compatibility of the conversion script with bin files seems nice to me. This might be a better fit for our (internal): https://github.com/tue-bmd/verascripts

wesselvannierop (commented):
Will review the rest later!

PR #540, opened 2026-08-03
Refactor directivity

wesselvannierop:
Small PR on top of #534 

Refactors the `directivity` func from simulator, such that pfield also uses it and there is no need for `_unnormalized_sinc` anymore. Small speed-up in the process.

## Summary by CodeRabbit

* **New Features**
  * Added a public ultrasound directivity calculation for modeling element response, including rigid and non-rigid baffle options.
  * Directivity calculations now support isotropic behavior when element width is unspecified.

* **Improvements**
  * Updated pressure-field and spectrum calculations to use frequency values in Hz.
  * Improved simulation efficiency by calculating sub-element directivity once and reusing it across frequency calculations.
  * Updated transmit and receive beamforming to use the shared directivity calculation.


vincentvdschaft (approved):
Nice improvement. Looks all good to me.

One thing, (maybe not for this PR though), but the simulator and the pfield module also both have a function to compute the spectrum of the pulse. Perhaps these could also be merged?

wesselvannierop:
> Nice improvement. Looks all good to me.
> 
> One thing, (maybe not for this PR though), but the simulator and the pfield module also both have a function to compute the spectrum of the pulse. Perhaps these could also be merged?

Maybe! Although from the code alone seems less obvious to me.

PR #541, opened 2026-08-03
Add a health check for the published zeahub datasets

tristan-deep:
Adding a simple lightweight check that checks the publicly maintained [datasets on zeahub](https://huggingface.co/zeahub/datasets). Note that this checks the status of the zeahub, not necessarily the code itself. Checks for number of files, size and zea version: 

- the files carry `zea_version >= 0.1.4`,
- the dataset still has the expected number of HDF5 files,
- the total size is roughly what we expect (±25%).

It reads HDF5 headers only, a couple of HTTP range requests per dataset, no data transfer. The full run is ~48 s over 12 datasets.

## Running it

```bash
pytest tests/data/test_zeahub_datasets.py -m heavy
```

It is marked `heavy`, so ordinary runs (`-m "not heavy"`) skip it and it lands in the existing heavy CI job rather than the main PR run.

## Summary by CodeRabbit

- **New Features**
  - Added an optional non-interactive approval mode for dataset uploads, allowing confirmations and missing revision creation to be accepted automatically.
  - Existing interactive confirmation behavior remains available by default.

- **Tests**
  - Added comprehensive health checks for maintained published datasets, validating file counts, approximate sizes, version metadata, and format compatibility.


Louisvh (dismissed):
Looks good, though the api monkey is maybe a bit of a false guarantee - I think the main way this code will break is by api changes in HF, which would break the code without breaking the test. Not really a problem and you wouldn't want to touch the real api in the tests; the curse of codecov enforcement, I suppose.

tristan-deep:
> Looks good, though the api monkey is maybe a bit of a false guarantee - I think the main way this code will break is by api changes in HF, which would break the code without breaking the test. Not really a problem and you wouldn't want to touch the real api in the tests; the curse of codecov enforcement, I suppose.

Yes fair point.. I now cut down on the tests that were anyways ignored with a `# pragma: no cover`.

Louisvh (approved):

PR #542, opened 2026-08-03
Fix `zlims` and `distance_to_apex` definition

tristan-deep:
Fixes a set of curved-probe geometry bugs that all trace back to one thing: `zlims` did not mean one thing.

`polar_pixel_grid` built its radii as `rlims = (zlims[0], zlims[1] + distance_to_apex)`. Only the far end got the apex distance added, so `zlims[0]` was a radius measured from the apex while `zlims[1]` was a depth measured from the transducer — two reference frames in one tuple, with nothing marking which was which. Callers had to guess, and several guessed wrong:

- `Parameters.coordinates_2d` applied `distance_to_apex` a second time on top of `rho_range`, which had already applied it. On a curvilinear probe that cropped the sector down to a strip near the far field.
- `Parameters.extent()` shifted only the near end for polar grids, stretching the depth axis.
- `Parameters.xlims`/`ylims` sized the sector from depth where they needed radius, under-reporting how wide a polar grid actually is.

Two further problems sat underneath. `distance_to_apex` was effectively hardcoded to zero: its schema default of `0.0` is written into `_params` by `BaseParameters.__init__`, so the fallback meant to compute it was unreachable, and no curved probe got a nonzero apex distance unless one was passed by hand. And the two halves of the pipeline disagreed on whether `zlims[1]` was in the grid at all — `polar_pixel_grid` placed radii with `np.arange` (far end excluded) while scan conversion read them back with `linspace` (included), so `rho_range` claimed the deepest sample sat a full sample deeper than it did. That stretched every scan-converted sector radially by `1/n`.

## The rule

Rather than patch each caller, this states the rule and applies it once. There are two frames:

| | `z = 0` at | used by |
|---|---|---|
| **Transducer frame** | the array surface | `zlims`, `Parameters.grid`, `Parameters.extent`, a scan conversion's reported `z_lim` |
| **Apex frame** | the point the beams fan out from | `rho_range`, the polar sampling coordinates |

`distance_to_apex` converts between them, applied in exactly the two places where a depth becomes a radius: `polar_pixel_grid`'s `rlims` and `Parameters.rho_range`. Nowhere else.

Both ends of `zlims` are now sample positions, matching `cartesian_pixel_grid`.

## What changed

**Geometry**

- `polar_pixel_grid` shifts both ends: `rlims = (zlims[0] + distance_to_apex, zlims[1] + distance_to_apex)`, and `Parameters.rho_range` matches.
- `radial_pixel_grid` counts its samples rather than walking a float range, so `rlims` are the radii of the first and last sample exactly. This retires the "trim in case of rounding errors" workaround.
- `Parameters.extent()` no longer shifts `zlims` for polar grids; `xlims`/`ylims` measure the sector from the apex.

**`Parameters.distance_to_apex`**

The apex is the point the beams fan out from, which for a curved array is the centre of curvature — so the apex distance is the probe's radius of curvature. It is now fitted from `probe_geometry` instead of guessed from the sector angle. Flat arrays give 0, an explicit value still wins, and the schema default is `None` meaning "fit it".

`fit_curved_probe_radius()` does the fitting: the inverse of `create_curved_probe_geometry()`, a least-squares circle fit that then checks itself against the elements, so a flat array or a geometry in another frame raises instead of returning the wrong circle. The tolerance is half an element pitch — loose enough for the irregularity of measured and simulated arrays, tight enough to catch a geometry off by an element.

**Scan conversion**

`compute_scan_convert_2d_coordinates`, `scan_convert_2d` and `ScanConvert` take `distance_to_apex` for one purpose: reporting `z_lim` as a depth below the transducer rather than from the apex, so it agrees with `Parameters.extent`. It never enters the sampling grid. The old argument of the same name did the opposite — it shifted `z_vec`, which crops the image instead of moving it.

**Terminology**

"Apex" now means one thing: the apex of the polar grid, i.e. the centre of curvature. The middle of a curved array is called the middle of the array. This ambiguity, raised in review, is what made the old docstrings hard to follow.

Asking a 3D grid for scan-conversion coordinates now reports what to do instead of raising a bare `NotImplementedError`.

## Migration

**`zlims[1]` is the deepest sample, not one past it.** A polar grid of `n` radial pixels spans `zlims` inclusively, so spacing is `(zlims[1] - zlims[0]) / (n - 1)`. If you chose `zlims` to hit a target spacing, drop one sample:

```python
zlims = (z_min, z_min + n * dr)        # before
zlims = (z_min, z_min + (n - 1) * dr)  # now
```

This has the widest reach: scan-converted output shifts for every polar grid, including those with `distance_to_apex = 0`, since the sector was previously stretched by one sample in `n`. Migrate `zlims` as above and the beamforming grid is bit-identical.

**Drop any manual apex shift.** If you pre-shifted `zlims[0]` to work around the old asymmetry:

```python
zlims = (distance_to_apex + z_min, z_max)  # before
zlims = (z_min, z_max)                     # now
```

**Curved probes on polar grids get a real apex distance.** `distance_to_apex` is not part of `ScanSpec` and is never stored in zea files, so it was `0.0` everywhere unless passed explicitly. A curved probe on a polar grid that does not set it now gets its fitted radius of curvature, changing the beamforming grid. That is the fix, but it will change output. Pass `distance_to_apex=0.0` to keep the old behaviour. Cartesian grids are unaffected either way.

**`scan_convert_2d(..., distance_to_apex=...)` now does the opposite.** It relabels `z_lim` and leaves the image alone. If you were passing it to shift the image, it was cropping it.

## Known limitation

`scan_convert_3d` has no `distance_to_apex` and places the apex at the origin. 3D polar grids are unsupported in `Parameters` regardless, and now say so.

## Summary by CodeRabbit

* **New Features**
  * Automatically estimates curved-probe apex distance from probe geometry.
  * Supports explicit or automatic apex-distance settings for polar grids and scan conversion.
  * Improved radial sampling includes requested endpoints and sample counts.
  * Added validation for unsupported or invalid probe geometries.

* **Documentation**
  * Clarified probe curvature, coordinate limits, radial endpoints, and scan-conversion behavior.

* **Bug Fixes**
  * Corrected depth and radial positioning for curved probes without altering converted image data.


zea/probes.py
```diff
@@ -204,6 +204,60 @@ def create_curved_probe_geometry(n_el, pitch, radius):
...
+
+    Elements are assumed to lie on an arc of radius R centred at ``(0, 0, -R)`` (apex at
+    the origin, zea's curved-probe convention -- see :func:`create_curved_probe_geometry`),
+    so ``x^2 + (z + R)^2 = R^2``  =>  ``x^2 + z^2 + 2 R z = 0``. Solved as a least-squares
+    fit over all elements rather than per-element, which is ill-conditioned for the
+    elements nearest the apex (``z -> 0``).
+
+    Args:
+        probe_geometry (np.ndarray): Element positions, shape (n_el, 3), in zea's
+            apex-at-origin curved-probe frame. Re-centre first if ``probe_geometry``
```
vincentvdschaft: I was a bit confused by this docstring because it mentions the apex being in the origin and also a `distance_to_apex`, which sounds like it should then be 0. If I understand correctly my confusion comes from there being two different types of apex: The apex of the curve (the middle), and the apex of the polar grid cone (the tip of the cone). Is my understanding correct? Perhaps this could be made a bit more clear

tristan-deep: I agree that is confusing, will rewrite a bit!

vincentvdschaft (dismissed):
This clarifies things a lot! It looks good to me.

zea/display.py
```diff
@@ -210,7 +209,7 @@ def compute_scan_convert_2d_coordinates(
         resolution = ops.mean([sRT, d_rho])  # mm per pixel
 
     x_vec = ops.arange(x_lim[0], x_lim[1], resolution)
-    z_vec = ops.arange(z_lim[0] + distance_to_apex, z_lim[1], resolution)
+    z_vec = ops.arange(z_lim[0], z_lim[1], resolution)
```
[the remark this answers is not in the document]
tristan-deep: Now have a more uniform definition across the board for zlims and distance_to_apex

tristan-deep:
> This clarifies things a lot! It looks good to me.

Thanks @vincentvdschaft ! Because of your comment I realized that actually definition for depth wise parameters (zlims and distance to apex) was not really uniform across beamforming (pixelgrid), scan conversion and parameter definitions. I added one more commit that fixes that. PR scope got extended a bit by that. I also updated the PR description.

tristan-deep:
@coderabbitai review

vincentvdschaft (approved):

tristan-deep:
As @vincentvdschaft suggested as a follow-up we can add a diagram to the docs that explains the geometry and links the variables.

PR #544, opened 2026-08-04
Add `t_peak` to simulator

vincentvdschaft:
Added `t_peak` to the simulator. It was not included before, leading to offsets in the beamformed images.

## Summary by CodeRabbit

* **New Features**
  * Added support for specifying transmit-pulse peak times in RF simulations.
  * Simulated RF timing now reflects the configured pulse peak for each transmit event.

* **Documentation**
  * Updated the simulation example and benchmark outputs to reflect the new timing behavior.
  * Removed an outdated warning from the recorded example output.


tristan-deep (approved):
Simpel but good! At some point we could do a more thorough check to see what computations / assumptions are missing in simulation, but is used in reconstruction, or the other way around.

PR #545, opened 2026-08-04
Fix matplotlib backend and opencv extras in the dev setup

### LINKED ISSUE #543: uv.lock is outdated
The current uv.lock does not include `pyqt6` so when one does `uv sync --all-extras --all-groups` after cloning the repo errors occur.

This happens for instance in the simulation notebook. The following will result in an error:
```bash
git clone https://github.com/tue-bmd/zea
cd zea
uv sync --all-extras --all-groups
uv run -m pytest "tests/test_notebooks.py::test_notebook_runs[zea_simulation_example.ipynb]"
```

I am not sure what the proper way is to upgrade the lockfile for zea. Can I just `uv sync` and then commit the generated lockfile?

tristan-deep:
Fixes #543. Two independent setup fixes.

## 1. Pin `MPLBACKEND=Agg` for the test suite

Notebook tests fail on a developer machine with:

```
ImportError: Failed to import any of the following Qt binding modules: PyQt6, PySide6, PyQt5, PySide2
```

Nothing to do with the lockfile — `pyqt6` isn't a dependency of anything here. The trigger is an interactive `MPLBACKEND` exported in the developer's shell, and two things let it through:

- `ipykernel` forces the inline backend, but only `if not os.environ.get("MPLBACKEND")` — an exported value survives into the papermill kernel.
- matplotlib's `backend_fallback` throws away an explicit interactive backend when there's no usable display. So headless machines and CI silently degrade to Agg and never see it; a workstation with a display does.

Setting `MPLBACKEND=Agg` in `conftest.py` before pyplot is imported fixes both the pytest process and every kernel it spawns. This replaces two weaker guards: `plt.rcParams["backend"] = "agg"` in `conftest.py`, which only affected the pytest process, and `os.environ.setdefault(...)` in `test_getting_started.py`, which was a no-op whenever a backend was already exported.

Not touching the notebooks: `Agg` makes `plt.show()` a no-op, and the pre-commit hook requires committed outputs, so the five notebooks that use it would silently lose their inline figures and the Colab links would render no plots.

## 2. Stop `display` and `display-headless` co-installing

Both extras provide `cv2`, so `uv sync --all-extras` installed both and kept whichever landed last. Uninstalling either then removed the shared `cv2/` directory, leaving the survivor recorded as installed but broken:

```
$ uv sync
Audited 196 packages in 1ms          # environment reported as fine
$ python -c "import cv2"
ModuleNotFoundError: No module named 'cv2'
```

Declaring the extras conflicting makes that an error at sync time instead. Two details worth knowing while reviewing:

- `dev` now depends on `opencv-python-headless` directly instead of `zea[display-headless]`. uv only activates a conflicting extra when it's named, so keeping the indirection left a plain `uv sync` with **no** cv2 at all.
- The second conflict pair (`display` vs `dev`) covers `--all-groups --extra display`, which would otherwise co-install both again through the dev group.

The `uv.lock` delta is large but mechanical: conflict markers only, no package added, removed, or changed version.

## Verification

Repro for #543, with `MPLBACKEND=QtAgg` and a virtual display:

```bash
MPLBACKEND=QtAgg xvfb-run -a uv run -m pytest \
  "tests/test_notebooks.py::test_notebook_runs[zea_simulation_example.ipynb]"
```
`1 failed in 97.71s` (PapermillExecutionError) before, `1 passed in 134.42s` after. `test_visualize.py`, `test_io_lib.py` and `test_generative.py` pass; full suite left to CI.

Every sync path, from a clean `.venv`:

| command | result |
|---|---|
| `uv sync` (contributing.rst) | headless cv2 |
| `uv sync --all-groups` | headless cv2 |
| `uv sync --all-extras` | error, names both extras |
| `uv sync --all-extras --all-groups` | error, names both extras |
| `uv sync --all-groups --extra display` | error, names extra + group |
| `uv sync --no-default-groups --extra display` | GUI cv2, for anyone who wants it |
| `uv sync --no-default-groups --group docs --extra jax` (readthedocs) | unchanged |
| `uv sync --frozen --inexact --dev` / `--no-dev` (Dockerfile) | unchanged |

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved headless test and development environments by ensuring plotting runs without requiring a display.
  * Prevented conflicting display configurations from being selected together.

* **Tests**
  * Standardized non-interactive plotting behavior across test processes.
  * Simplified test setup and removed redundant environment configuration.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -111,13 +111,24 @@ docs = [
     "PyStemmer",
     "IPython",
 ]
-# The full dev environment: the tooling groups plus the runtime extras
+# The full dev environment: the tooling groups plus the runtime extras. Takes
+# opencv-python-headless directly, not via `zea[display-headless]`: uv only activates a
+# conflicting extra when it is named, so the indirection would leave `uv sync` without cv2.
 dev = [
     { include-group = "tests" },
     { include-group = "docs" },
     { include-group = "lint" },
-    "zea[display-headless]",
+    "opencv-python-headless>=4",
     "zea[models]",
+]
+
+# Both extras provide `cv2`, so co-installing them leaves whichever landed last, and
+# uninstalling either takes the shared `cv2/` with it. `dev` already brings the headless
+# build, hence the second pair.
+[tool.uv]
+conflicts = [
+    [{ extra = "display" }, { extra = "display-headless" }],
+    [{ extra = "display" }, { group = "dev" }],
 ]
 
 [tool.hatch.build.targets.wheel]
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -7,13 +7,8 @@ import sys
 from collections import defaultdict
 from pathlib import Path
 
-# Pin the matplotlib backend for the whole suite, before pyplot is imported anywhere in
-# this process. An interactive MPLBACKEND exported in a developer's shell is otherwise
-# inherited by the papermill kernels in test_notebooks.py, where it survives ipykernel's
-# ``if not os.environ.get("MPLBACKEND")`` guard and fails the notebook with "Failed to
-# import any of the following Qt binding modules". matplotlib's backend_fallback hides
-# this on headless machines and in CI, so it only shows up on a workstation with a
-# display.
+# Must be set before pyplot is imported, and inherited by the papermill kernels in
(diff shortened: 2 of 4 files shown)


tristan-deep:
@vincentvdschaft could you check your setup again, but on this branch. I think it would also resolve your issues (#543).

PR #550, opened 2026-08-05
Factored common terms in tx and rx path of the simulator (speedup 30~600x)

Louisvh:
Currently, the simulator is _nearly_ factorizable into tx->scatterer and rx->scatterer paths, except for the spread term (which contains 1/(dist_rx + dist_tx), meaning it can't factor). That means the whole thing has to compile to a giant nested loop over all tx elements, all rx elements, all frequencies and all scatterers, with a lot of the computations being done in the deepest layer. By approximating the spread term with the geometric mean (1/(2sqrt(dist_rx × dist_tx))), everything separates nicely, and it no longer needs to construct a giant element^2 × frequencies matrix for every scatterer. Speedup is dependent on settings, but it's anywhere between 30x and 600x in my tests.

The spread approximation isn't perfect, but it's pretty close. Even in the worst case (big aperture, close to the probe), I'm getting a near-perfect match with the original path:

image

This branch also includes Vincent's time-domain version, which is much faster in some cases, but slower in others:

linear (80 el), 2338 samples, 2049 bins, seconds per transmit (stopped if previous entry > 4s)

scatterers →           |     512   |    2048  |    8192   |   32768    |  131k   |   524k  |  1M
---------------|-----------|----------|-----------|------------|-----------|-----------|---------
exact      |   0.053   |   0.230  |   1.797   |   4.702    |       -   |        -  |        -
factored   |   0.007   |   0.009  |   0.039   |   0.198    |   0.642   |    2.553  |    5.128
fast           |   0.004   |   0.004  |   0.011   |   0.030    |   0.116   |    0.590  |    0.970

matrix (32×32 el), 2338 samples, 2049 bins, seconds per transmit (stopped if previous entry > 4s)

scatterers →     |     512   |    2048   |    8192  |    32768   |  131k
-----------|-----------|-----------|----------|------------|--------
exact      |  15.310   |       -   |       -  |        -   |       -
factored   |   0.025   |   0.109   |   0.404  |    1.534   |   6.092
fast       |   0.133   |   0.477   |   2.087  |    7.572   |       -

----

Note: the fish images are changes that were missing pieces of #544 ; re-running the notebook on current `main` produces a fish RF with a nonzero `t_peak`. I've updated the images so the readthedocs page doesn't drift.

## Summary by CodeRabbit

- **New Features**
  - Added exact, factored, fast, and time-domain ultrasound simulation modes.
  - Added configurable simulator method selection, with factored simulation as the default.
  - Added pulse waveform generation and support for batched and unbatched simulations.

- **Bug Fixes**
  - Improved simulation accuracy, performance, and handling of scatterer geometry and signal generation.

- **Tests**
  - Added regression coverage for scatterer visibility and image-quality thresholds across simulator modes.

- **Documentation**
  - Updated benchmark results and expanded API documentation for simulation features.


tristan-deep:
I see this is mainly missing some tests @Louisvh. Let me know can add this to [v0.1.5 milestone](https://github.com/tue-bmd/zea/milestone/18) of this Friday or defer to v0.1.6. Very cool update, crazy speedups.

Louisvh:
Yes, feel free to add it; I have blocked some time tomorrow for working on it.

Louisvh:
While finishing this up, I realized the calculation of the spread in the original path was incorrect (and that mistake propagated to the factored version): the spread is being calculated as $1/(d_{tx}+d_{rx})$, but that implies the scatterer doesn't restart the spreading. The returning wave keeps diverging as if it's still the same wavefront instead of re-starting the spread pattern (which would be $1/(d_{tx} \cdot d_{rx})$ ), so it's behaving like a mirror instead of like a scatterer:

spreading_model_beam

This effect was partially hiding behind the fact that the elevation lens wasn't modeled, so in a 2D-simulation, everything was close to correct; amplitude decayed approximately a factor `r` too little via the return spread, but a factor `sqrt(r)` too much via the transmit spread.

The good news is that this makes the exact path factorizable, so the three simulator modes drop back to two fast ones. The bad new is that this PR just became a lot harder to review. I've pushed an updated/cleaned up version that collapses the `factored` path into the `exact` one with the updated spread.

Louisvh:
I'll split the updated spread function off into a separate PR. That way, we'll keep a faster version of the legacy simulator available in the git tree (so we can more easily reproduce old experiments). The spread is a big breaking change, so better to keep that PR small and review-able.

zea/simulator.py
```diff
@@ -226,6 +280,310 @@ def simulate_rf(
     return rf_data
 
 
+def simulate_rf_fast(
```
vincentvdschaft: This simulator currently outputs channel data with different scaling from the other simulator. We might want to make the scaling uniform or at least warn about this?
| Simulator | Min | Max | Std |
|---|---|---|---|
| Factored | -0.0030091414 | 0.0027273507 | 0.00021771448 |
| Unfactored | -0.0030089242 | 0.0027272592 | 0.0002176403 |
| Fast | -0.040072076 | 0.03553158 | 0.0028682074 |

zea/simulator.py
```diff
@@ -173,51 +170,108 @@ def simulate_rf(
...
+                sampling_frequency=sampling_frequency,
+            )
 
-        result = (
-            waveform_spectrum[None, None, None]
-            * delay2(
-                freqs[None, None, None],
-                tau_total[..., None],
+            # When each element fires, relative to the start of the recording. [n_txel, 1]
+            tx_firing_times = t0_delays[tx][:, None] - initial_times[tx] + t_peak[tx]
```
vincentvdschaft: I think this name is a bit confusing. The variable name suggests that this would just be the t0_delays. I am not sure what would be better. Perhaps something like `shifts_not_travel_related`?

zea/simulator.py
```diff
@@ -226,6 +280,310 @@ def simulate_rf(
...
+    center_frequency,
+    sampling_frequency,
+    t0_delays,
+    initial_times,
+    element_width,
+    attenuation_coef,
+    tx_apodizations,
+    t_peak,
+):
+    """Time-domain (splat-and-convolve) RF simulator.
```
vincentvdschaft: I think it may be nice to move this simulate function and its helpers to a separate module to make this file less overwhelming. What do you think?

vincentvdschaft:
This is a great improvement! This will probably help a lot with making the tests faster too. I placed a few comments, but other than that it looks good to me.

zea/simulator.py
Louisvh: Yes, that's probably better. I previously renamed it to "delay_offsets", but `_not_travel_related` is clearer.

zea/simulator.py
Louisvh: Yeah, agreed. There's quite little shared code, and a lot of lines to scroll through. Easy to get lost what is part of which one, too. I've split them off now, with `simulator.py` being the "main" file where `simulator_time_domain.py` imports a few functions from. I've renamed _fast to _time_domain, to make it clearer what the difference is (also because it's not actually faster for very large numbers of elements due to the `n_el**2` term that I don't think we can get rid of there).

zea/simulator.py
Louisvh: Yeah, it's because the frequency-domain version was scaling the pulse depending on `n_fft` and the sampling frequency. I've normalized that one - now they're in the same range again (and don't magically change amplitudes if you change seemingly unrelated parameters).

vincentvdschaft (approved):

PR #553, opened 2026-08-07
Add polar pixelgrid image and explanation to docs

vincentvdschaft:
Created an image and small explanation of the limits variables defining the image grid in the docs as discussed in
- #542

This can hopefully help people understand the relation between `rlims`, `zlims`, and `distance_to_apex` better.

## Summary by CodeRabbit

* **Documentation**
  * Added guidance for defining beamforming grids.
  * Documented Cartesian and polar grid extents, including distance from the aperture apex.
  * Added responsive illustrations with captions for polar pixel-grid geometry in dark and light modes.


tristan-deep:
Very nice diagram! Super clear. 

One comment, I see it is now added to the main Parameters page, [here](https://zea--553.org.readthedocs.build/en/553/config.html#defining-a-beamforming-grid). This seems a bit detailed for that page, maybe closer to the page where this parameters are actually listed. However, on closer look of the [Parameters page](https://zea--553.org.readthedocs.build/en/553/config.html), that also seems a bit of a mess itself... I don't think it links to the [Parameters API reference](https://zea--553.org.readthedocs.build/en/553/_autosummary/zea.Parameters.html) which does include a description of all parameters.

wesselvannierop:
It does tie together a couple of params, so this might be the right place?

tristan-deep:
> It does tie together a couple of params, so this might be the right place?

Yes upon further thought I do think it belongs here. We just need better layout of that page as it feels a bit random. But that we can make an issue, let’s do that @coderabbitai

tristan-deep (approved):

PR #555, opened 2026-08-07
Drop the batch axis from posterior_sample

### LINKED ISSUE #111: Suggestion to remove batch dim from DiffusionModel.posterior_sample
The logic here stacks the `n_samples` dim with the `batch_size`. However, another `Operator` might need another key as `mask`. I propose to remove the batch dimension all together, keeping the zea implementation simple. If you really want to batch posterior sampling, you can just use `ops.map` or `ops.vectorized_map`.

https://github.com/tue-bmd/zea/blob/86bf92262fd278de9fb509a647087cf61474ad7e/zea/models/diffusion.py#L238-L248

wesselvannierop:
Closes #111

This PR aligns the API of `sample()` and `posterior_sample()`. Both operate on `(n_samples, *input_shape)`. You can use `zea.func.vmap` to operate on batches of posterior samples, like is done in the example notebooks ([diffusion](https://github.com/tue-bmd/zea/blob/fix/diffusion-posterior-sample/docs/source/notebooks/models/diffusion_model_example.ipynb), [HVAE](https://github.com/tue-bmd/zea/blob/fix/diffusion-posterior-sample/docs/source/notebooks/models/hvae_model_example.ipynb)). This means `posterior_sample` no longer has to magically batch all its arguments (such as `mask`, for inpainting); that is now up to the user.

**Breaking ⚠️:** `posterior_sample()` and `HierarchicalVAE.partial_inference()` used to return `(batch_size, n_samples, *input_shape)`, and `DiffusionModel`'s `initial_samples` used to take a batch axis. Batched callers should wrap the call in `vmap`. Measurements may still be passed as `(n_samples, *input_shape)` to condition each sample on a different measurement.

Additionally, adds a seed argument to HVAE's sampling methods, which reproduces the previous behavior on main, while using `vmap`. It also means the methods can now be used themselves in a traced context without duplicating seeds.

## Summary by CodeRabbit

- **New Features**
  - Posterior sampling now supports shared or per-sample measurements across generative models.
  - Optional seeds enable reproducible, independent sampling across workflows.
  - Sampling returns flattened batches and preserves all generated reconstructions.

- **Bug Fixes**
  - Improved measurement, mask, and tensor shape validation.
  - Visualization and inference now process complete posterior sample sets.
  - Conditional sampling produces independent per-measurement results where applicable.

- **Documentation**
  - Updated model examples and notebooks to reflect revised sampling behavior and output shapes.


wesselvannierop:
@tristan-deep @swpenninga codecov fails because my patch of the hvae now triggers codecov to see it as part of this PR

tristan-deep:
> @tristan-deep @swpenninga codecov fails because my patch of the hvae now triggers codecov to see it as part of this PR

Yes that is a bit annoying... Currently the zea.models is the only subpackage with poor coverage, and hvae especially. See [codecov](https://app.codecov.io/gh/tue-bmd/zea). So it is very likely to hit low patch coverage when you touch those files.

zea/models/hvae/__init__.py
```diff
@@ -106,83 +107,124 @@ def custom_load_weights(self, preset, load_weights=True, **kwargs):
...
-        but does n_samples decoder passes to create posterior samples.
+        Performs posterior sampling for a single measurement.
+
+        The encoder is deterministic, so one measurement needs only a single
+        encoder pass; the n_samples decoder passes are what produce distinct
+        posterior samples.
 
         Args:
-            measurements (tensor): Input measurements of shape [B, 256, 256, 3].
+            measurements (tensor): Input measurements of shape [256, 256, 3] in
```
tristan-deep, line 137: Might be nice to add the vmap instructions here in the docstring of sample and posterior sample so they know how to run on a batch efficiently? Probably for all generative models?

tristan-deep:
I started a job to improve codecoverage for zea.models. Will create a PR for that which can be merged before this one and that should fix the patch coverage here as well.

zea/models/hvae/__init__.py
```diff
@@ -106,83 +107,124 @@ def custom_load_weights(self, preset, load_weights=True, **kwargs):
...
+        measurements = ops.convert_to_tensor(measurements)
+        if len(measurements.shape) != 4:
+            return measurements[None], False
+
+        # A singleton leading axis is shared by every sample, just like no axis.
+        batch_size = measurements.shape[0]
+        if batch_size == 1:
+            return measurements, False
+        if batch_size != n_samples:
+            raise ValueError(
```
tristan-deep: A warning like this would also be nice for e.g., diffusion and flow matching generative models. Maybe can add to the base class so it applies for all?

zea/models/hvae/__init__.py
wesselvannierop: Sure, but I'd just propose a link to the `vmap` function.

zea/models/hvae/__init__.py
wesselvannierop: I'll have a look!

tristan-deep:
> I started a job to improve codecoverage for zea.models. Will create a PR for that which can be merged before this one and that should fix the patch coverage here as well.

See #560

tristan-deep (approved):

Issue #318, opened 2026-04-16
OpenH-RF: Open Data Initiative 🧩  


tristan-deep:
We are proud to collaborate and support the [OpenH-RF](https://github.com/open-h/OpenH-RF) initiative with `zea`. See a description of OpenH-RF below. 

> OpenH-RF is a collaborative initiative led by [Stanford University](https://med.stanford.edu/ultrasound.html), [Eindhoven University of Technology](https://www.tue.nl/en/research/research-groups/signal-processing-systems/biomedical-diagnostics) and [NVIDIA](https://www.nvidia.com/) to build a large-scale, openly licensed dataset of pre-beamformed (channel capture) medical ultrasound measurements. The goal: train general-purpose foundation models capable of multi-task raw-to-insight inference across echocardiography, general, fetal and transcranial imaging, blood flow measurement and ultrasound inverse problems.

> We aim to curate 20,000+ real and synthetic channel capture measurements spanning reconstruction, flow, quantitative imaging, motion estimation and interpretation tasks — released under CC BY 4.0.

<img width=100% alt="Image" src="https://github.com/user-attachments/assets/13814fdc-0b69-498f-a2f9-e29559e0110e" />

---

`zea` will contribute in the following ways:
- Provide a complete and standardized data specification for the OpenH-RF dataset, through [`zea.File`](https://zea--307.org.readthedocs.build/en/307/data-acquisition.html) and [`zea.Dataset`](https://zea.readthedocs.io/en/latest/_autosummary/zea.data.datasets.html#zea.data.datasets.Dataset) #307 
- Enable automated dataset discovery and downloading via `zea.File`, leveraging the Hugging Face API.
- Define reproducible ultrasound signal processing pipelines ([`zea.Pipeline`](https://zea.readthedocs.io/en/latest/pipeline.html)) using configuration files stored alongside the datasets. 
- Provide data and training infrastructure required to develop the first ultrasound radio-frequency channel foundation model.

tristan-deep:
The [request for proposals](https://github.com/open-h/OpenH-RF#how-to-participate) is now closed! 

For all submitters, have a look at our [latestrelease](https://github.com/tue-bmd/zea/releases/tag/v0.1.0) which offers lots of improvements to file format and creation. Stay tuned for more releases, these will always include general improvements and none-breaking changes to the file format. Meaning that you can start with submission without having to worry your script breaks.

PR #556, opened 2026-08-07
Introduction of torch.compile()

swpenninga:
Previously we did not support JIT compilation for torch, so here I added it to the zea jit functionality.
This had implications across notebooks, the simulator, pfield, beamformer, autograd, and the tests. 
In this PR there are many 1-line changes that are equal in functionality, but required for the torch backend.
I also added g++ to the Dockerfile for the C++ compiler of torch.

## Summary by CodeRabbit

- **New Features**
  - Added PyTorch support for JIT compilation through `torch.compile`.
  - Added backend-agnostic JIT guidance and controls for enabling or disabling compilation.
  - Improved serialization of compiled functions and methods, including stable handling across compilations.

- **Bug Fixes**
  - Enabled gradient and value calculations for compiled PyTorch functions.
  - Improved validation for ultrasound band-pass ranges.
  - Fixed numerical handling in simulation and waveform calculations.

- **Documentation**
  - Updated examples to document PyTorch compilation and optional compilation warm-up behavior.


zea/beamform/pfield.py
```diff
@@ -238,15 +238,12 @@ def probe_spectrum(f):
...
-    attenuation_wavenumber = ops.cast(attenuation_wavenumber, dtype="complex64")
 
     # Exponential array for the increment wavenumber dk
     wavenumber_step = 2 * np.pi * freq_step / sound_speed
     attenuation_wavenumber_step = attenuation_coef * freq_step
-    wavenumber_step = ops.cast(wavenumber_step, dtype="complex64")
-    attenuation_wavenumber_step = ops.cast(attenuation_wavenumber_step, dtype="complex64")
 
-    @jit
+    @jit(torch=False)
```
OisinNolan, line 246: Does this mean that we can't jit pfield computation with torch backend? Curious what the error was here!

OisinNolan (commented):
Nice work! Just left a small question. Will take a look again when codecov passes

zea/beamform/pfield.py
swpenninga: Indeed, torch.compile() crashes for the pfield computation, where JAX and TF work fine. 
Apparently it does not like "loop-carried complex buffers". 
I wasn't comfortable editing the functionality of the pfield computation in this PR so this is here for now..

tristan-deep: @coderabbitai create a separate issue for pfield jit compatibility with torch backend

tristan-deep (changes requested):
Nice!!

Tested this for the Robeaute [OpenH-RF](https://github.com/tue-bmd/zea/issues/318) submission, only had to apply [8e9a9c0](https://github.com/tue-bmd/zea/pull/556/commits/8e9a9c0e777239dd47f9aabf8ae8ecaf7b7134e8) and it did work! 

I do get a mega warning and log stream, but that might just be expected with `torch.compile`? Also interesting part of that was: `UserWarning: Torchinductor does not support code generation for complex operators. Performance may be worse than eager.`

EDIT: was trying to run the pipeline notebook example now (using a default bmode pipeline) but that seems to fail. Any idea why? I think at least supporting this usecase would be nice. Of course we don't have to support all sorts of pipelines. 
EDITEDIT: @swpenninga I was wrong, had to rebuild the docker env to get the right installations and now the notebook runs fine!

zea/models/diffusion.py
```diff
@@ -1265,7 +1267,7 @@ def __call__(
             A ``(gradients, (measurement_error, (pred_noises, pred_images)))``
             tuple (see :meth:`call`).
         """
-        return self.call(
+        return self._call(
```
tristan-deep, line 1270: what happens here?

tristan-deep (dismissed):

zea/models/diffusion.py
swpenninga: Long story:
```
# old
self.call = jit(self.call)   # overwrites the public `call` method on the instance

# new
self._call = self.call if self.disable_jit else jit(self.call)  # stashed separately
```
torch.compile-wrapped callables arent picklable. self.call = jit(self.call) puts a compiled object into self.__dict__ and any attempt to serialize fails.

swpenninga: So any model you would try to save with torch.compile() would fail to load if you didn't do this, the old ones now still have self.call to load

tristan-deep (approved):
@swpenninga good to go? thanks for the update, useful to have! Especially for [Windows users](https://zea.readthedocs.io/en/latest/installation.html#windows-users).

PR #558, opened 2026-08-10
Dropping `zea.internal.core.Object`

wesselvannierop:
zea.internal.core.Object had no clear purpose, so it is dropped.

- serialized and the item accessors move to BaseParameters; merge and the pytree helpers
  had no call sites, and Operator/DiffusionGuidance used none of it.
- BaseParameters compares by content and is unhashable. Object's __eq__/__hash__ pickled
  the whole __dict__, which raises for any subclass owning a Keras model or a closure and
  broke tf.function tracing.
- serialized no longer caches its checksum, which went stale on in-place mutation, and
  now includes the class name. Invalidates existing ~/.cache/zea entries once.
- serialize_elements picks up anything exposing a serialized property.

Ideally merged before #555 

## Summary by CodeRabbit

* **New Features**
  * Added content-based equality and dictionary-style access for parameter objects.
  * Improved serialization for custom objects and parameter configurations, including type-aware cache keys.
  * Expanded support for callable, tensor, JSON, and wrapper serialization scenarios.

* **Bug Fixes**
  * Improved cache behavior after parameter updates, deletions, and invalidation.
  * Prevented unprepared parameter objects from being passed into pipelines.
  * Fixed TensorFlow tracing compatibility for diffusion sampling.

* **Refactor**
  * Simplified the object model and removed the legacy shared base class.


wesselvannierop:
Also adding @benluijten because I think he made this functionality at some point

zea/internal/core.py
```diff
@@ -93,14 +93,6 @@ def __setattr__(self, name: str, value):
             self._serialized = None
         super().__setattr__(name, value)
 
-    def __eq__(self, other):
-        if not isinstance(other, self.__class__):
-            return False
-        return self.serialized == other.serialized
-
-    def __hash__(self):
```
tristan-deep, line 101: hash removed from Object but not added to Parameters. Are we loosing functionality?

wesselvannierop: `__hash__` is what lets an object be a dict key/set member; defining `__eq__` without it makes a class unhashable. The branch made BaseParameters unhashable and dropped content == from Object. Nothing in the repo hashes these objects (the cache keys off serialized, not __hash__), and a content hash on a mutable object would be unsound anyway.

wesselvannierop:
I think I am going to drop the `Object` class, it doesn't serve a clear purpose, and is inherited by classes that don't even use its functionality. Working a bit more on this PR.

tristan-deep (dismissed):
quite technical diff, but conceptually I agree and see no standout things happening in the diff. Would be good to wait for the rabbit review, but considering all tests pass seems good to go!

tristan-deep:
@coderabbitai full review

benluijten:
Given the recent refactor and changes to the Parameters class, I think it makes sense to remove the zea Object. 

The equality check was added such that we can quickly check if there are any updates to e.g. the Scan class, such that we can determine whether to recompute a function or not. Could make a big performance difference in a loop. 
I believe this is now mostly addressed inside the Parameters class (e.g. update() doesnt trigger a full recompute), but having a fast equality check is still quite usefull.
I did not do a full review yet, but comparison speed shouldnt be significantly affected by this change right? Similar hashing strategy under the hood?

zea/internal/core.py
```diff
@@ -86,7 +98,7 @@ def dict_to_tensor(dictionary: dict, keep_as_is: list | None = None) -> dict:
         # Get the value from the dictionary
         value = dictionary[key]
 
-        if hasattr(value, "to_tensor"):
+        if _supports_zea_to_tensor(value):
```
wesselvannierop: I'd rather check explicitly for the zea classes that have to_tensor.

tristan-deep: that makes sense actually

wesselvannierop:
> Given the recent refactor and changes to the Parameters class, I think it makes sense to remove the zea Object.
> 
> The equality check was added such that we can quickly check if there are any updates to e.g. the Scan class, such that we can determine whether to recompute a function or not. Could make a big performance difference in a loop. I believe this is now mostly addressed inside the Parameters class (e.g. update() doesnt trigger a full recompute), but having a fast equality check is still quite usefull. I did not do a full review yet, but comparison speed shouldnt be significantly affected by this change right? Similar hashing strategy under the hood?

Yes, the same logic just moved to `Parameters`. It doesn't cache the serialized property anymore though, as that is unsafe for in-place modification. Shouldn't be too much slower, but for pure speed under the guarantee no in-place updates are done, this could be cached. Also does not implement __hash__ anymore.

tristan-deep (dismissed):

tristan-deep (dismissed):

tristan-deep (approved):

PR #566, opened 2026-08-13
Fixed the spread model of the simulator and added an elevation lens option

Louisvh:
While working on #550, I realized the calculation of the spread in the original path was incorrect (and that mistake propagated to the factored version): the spread was being calculated as $1/(d_{tx}+d_{rx})$, but that implies the scatterer doesn't restart the spreading. The returning wave keeps diverging as if it's still the same wavefront instead of re-starting the spread pattern (which would be $1/(d_{tx} \cdot d_{rx})$ ), so it was behaving like a mirror instead of like a scatterer:

spreading_model_beam

This effect was partially hiding behind the fact that the elevation lens wasn't modeled, so in a 2D-simulation, everything was close to correct; amplitude decayed approximately a factor `r/2` too little via the return spread, but a factor `sqrt(r)` too much via the transmit spread. This PR fixes that, and adds an `elevation_lens` option to swap between a spherical and a cylindrical transmission model.

The good news is that this makes the exact path factorizable, so the three simulator modes drop back to two fast ones. The bad new is that this breaks compatibility in various ways. The old version resulted in RF that was similar to real RF with a (slightly wonky) TGC-curve. However, the simulator doesn't actually support TGC, so the corrected version results in worse-looking images. We may want to add a simple auto-TGC before we merge this one.

## Summary by CodeRabbit

* **New Features**
  * Added elevation-lens simulation support with configurable element height and memory-aware scatterer chunking.
  * Added automatic filtering and bucketing of out-of-plane scatterers for improved simulation efficiency.
  * Introduced updated simulation methods: exact, frequency approximation, and time approximation.
  * Improved pulse scaling and propagation behavior for elevation-lens simulations.

* **Bug Fixes**
  * Improved handling of empty scatterer sets and record-length edge cases.
  * Prevented pruned scatterers from affecting simulation outputs.


Louisvh:
The failing test is a latent bug, surfacing because of louder simulator output. Fixed by #565.

Louisvh:
I've un-stacked this PR from #550, because stacking would squash both into one main merge commit. Part of the reason for splitting them is to preserve a fast version of the legacy simulator, which only remains in the tree if they're separate merges to main.

tristan-deep:
> I've un-stacked this PR from #550, because stacking would squash both into one main merge commit. Part of the reason for splitting them is to preserve a fast version of the legacy simulator, which only remains in the tree if they're separate merges to main.

Ah fair point! Didn't know stacking would squash the commits, seems like kind of a downside indeed...

Louisvh:
Correction, it apparently doesn't! I just saw a button with "Squash and merge stack", assuming that would, in fact, squash and merge the stack. But [it actually squashes and merges the PRs as separate commits](https://docs.github.com/en/pull-requests/reference/stacked-pull-requests#merge-methods); it doesn't squash the stack.

tristan-deep:
> Correction, it apparently doesn't! I just saw a button with "Squash and merge stack", assuming that would, in fact, squash and merge the stack. But [it actually squashes and merges the PRs as separate commits](https://docs.github.com/en/pull-requests/reference/stacked-pull-requests#merge-methods); it doesn't squash the stack.

Okay awesome, that is much more useful! First stack of zea. Would've been nice to have this for #358 😅

tristan-deep:
If I understand correctly, @Louisvh you are planning to add one more PR to this stack which implements the auto TGC? That way all simulator PRs get merged at the same time and will arrive at main together.

Louisvh:
> If I understand correctly, @Louisvh you are planning to add one more PR to this stack which implements the auto TGC? That way all simulator PRs get merged at the same time and will arrive at main together.

Yup! Now added to the stack → #573

zea/simulator.py
```diff
@@ -136,132 +158,53 @@ def simulate_rf(
     peak = ops.max(ops.abs(pulse))
```
vincentvdschaft: This normalization might be nicer to do in get_pulse_spectrum_function. There is already a hann_unnormalized that can be used I think.

zea/simulator.py
```diff
@@ -104,9 +112,23 @@ def simulate_rf(
 
     element_width = _resolve_element_width(probe_geometry, element_width)
 
+    if element_height is None:
+        element_height = element_width
+
+    magnitudes = scatterer_magnitudes
+    if elevation_lens:
```
vincentvdschaft, line 118: This is only for 2d right? We may want to warn if used with a 3d array or mention that it is 2d only in the docstring.

zea/simulator_time_domain.py
```diff
@@ -197,11 +233,13 @@ def _precompute_scatterer_response(
         scatterer_positions, probe_geometry, element_width, sound_speed, center_frequency
     )
     directivity_pair = element_directivity[:, :, None] * element_directivity[:, None, :]
-    spread_attenuation = spread(two_way_distance) * attenuate(
-        center_frequency, attenuation_coef, two_way_distance
+    spread_attenuation = (
+        spread(one_way_distance[:, :, None], 0.5 if elevation_lens else 1.0)
```
vincentvdschaft, line 241: I think this makes sense, but it needs to be explicitly mentioned in the docs somewhere that elevation_lens has this effect (not only filtering scatterers, but also assuming energy spread is 2d).

zea/simulator.py
```diff
@@ -136,132 +158,53 @@ def simulate_rf(
     peak = ops.max(ops.abs(pulse))
     waveform_spectrum = nonnorm_spectrum / ops.cast(peak, "complex64")
 
+    scat_pos_relative_to_probe = scatterer_positions[:, None] - probe_geometry[None]
+    theta = ops.arctan2(scat_pos_relative_to_probe[..., 0], scat_pos_relative_to_probe[..., 2])
+    phi = ops.arctan2(scat_pos_relative_to_probe[..., 1], scat_pos_relative_to_probe[..., 2])
+
+    # [n_scat, n_el, n_freq]
+    directivity_x = directivity(freqs[None, None], theta[..., None], element_width, sound_speed)
+    directivity_y = directivity(freqs[None, None], phi[..., None], element_width, sound_speed)
```
vincentvdschaft: If we are introducing element_height, then I guess this should be element_height right? Although for the very large element_height for elevation focusing this is not very accurate anymore (we are more in the near field of the element then). I am not sure how to handle this actually. Do you have any thoughts about this? Perhaps just mentioning this somewhere is fine?

zea/simulator_time_domain.py
Louisvh: Now added to the docstring!

zea/simulator.py
Louisvh: Yup, should be height, fixed now. I think it's fine in context of the very simple elevation lens model; we roll off too much energy via the directivity, but we don't roll off enough via the lens, so it kind of balances out.

zea/simulator.py
Louisvh: There's now a check whether all elements are on the same elevation, and it logs 

```python
log.warning(
 "elevation_lens=True models a 1D probe with a cylindrical lens, but the probe is not"
 f"1D (element elevation min, max: {elevation.min()}, {elevation.max()}) "
 "This is probably a mistake."
)
```


if not. We could let it crash instead, but I figured there may be situations where you'd actually want this (e.g. for something similar to the per-element-dome-lens probe you played around with a while ago).

This also warns on a rotated 1D probe, which also isn't supported, so the warning isn't quite universal (but it will still correctly point a user towards the problem, at least).

vincentvdschaft: Nice!

vincentvdschaft (dismissed):

Louisvh:
The stack contained a manually fixed merge conflict, which stacks apparently don't handle well yet. I think what happens is it first squashes the first PR, forgetting all about the merge resolution (which refers to commit hashes that no longer exist after the squash), and when it tries to squash and replay the next PR, it hits the now-unresolved conflict, failing to merge. I've now merged the first PR separately and re-applied the merge resolution manually with `main` as parent. The others should now apply correctly.

vincentvdschaft (approved):

PR #570, opened 2026-08-17
Added REFoCUS for IQ

RobinVV:
REFoCUS ops now interprets data as RF or IQ from n_ch in data.
This allows REFoCUS to operate correctly on IQ data.

## Summary by CodeRabbit

* **New Features**
  * Refocusing workflows now support both RF and IQ signal decoding.
  * IQ processing supports configurable demodulation-frequency offsets.
  * Refocusing validates supported one- and two-channel inputs across batched and unbatched workflows.

* **Bug Fixes**
  * Corrected ramp-filter scaling for negative-frequency IQ data to prevent sign inversion.
  * Added validation requiring a demodulation frequency for IQ input.


zea/ops/pipeline.py
```diff
@@ -2091,14 +2091,14 @@ def _get_hinv(self, delays, f_vec, apod):
...
         """REFoCUS decoding for a single (unbatched) volume.
 
         All channels and all frequency bins are processed in parallel via
         batched tensor operations.
 
         Args:
-            data: ``(n_tx, n_ax, n_el, n_ch)`` float32 RF array.
+            data: ``(n_tx, n_ax, n_el, n_ch)`` float32 array.
             delays_samples: ``(n_tx, n_el)`` transmit delays in samples.
             apod: ``(n_tx, n_el)`` transmit apodization.
```
swpenninga, line 2236: Can you add descriptions of the new inputs here as well? (in the same format, so they will be visible in the `read_the_docs`).

zea/ops/pipeline.py
```diff
@@ -2091,14 +2091,14 @@ def _get_hinv(self, delays, f_vec, apod):
         sinv_c = ops.cast(sinv, "complex64")
         return ops.matmul(VHT * sinv_c[:, None, :], UT)
 
-    def _decode(self, data, delays_samples, apod):
+    def _decode(self, data, delays_samples, apod, demodulation_frequency, sampling_frequency):
```
swpenninga, line 2221: These arguments will not be present if demodulation did not happen yet (RF version), right? Will this not crash?

swpenninga:
all 15 test failures are in `tests/test_refocus.py`. This is because they do not have a demodulation_frequency that is now required, but this makes sense, they have not been demodulated yet. I think this should be made into an optional argument (and it should be read from the pipeline automatically)

zea/ops/pipeline.py
swpenninga: [here](<https://zea--570.org.readthedocs.build/en/570/_autosummary/zea.ops.html#zea.ops.Refocus>)

swpenninga:
Note to self: requires checking the ramp-filtered adjoint also

tristan-deep:
@swpenninga, @RobinVV ready for merge?

swpenninga:
Don't merge it yet please

swpenninga:
Added a small fix for the jit compiled torch version, which broke down.
Also expanded the test suite to make sure IQ REFoCUS always equals RF REFoCUS, before that it was only shape testing.

swpenninga (approved):

PR #573, opened 2026-08-18
Added noise and TGC options to simulator

Louisvh:
This adds `noise_level_db` and `tgc_max_db` options to the simulator, so you can now upgrade your fish to a noisy fish. Noise is white noise added to the raw trace relative to the trace max (or relative to an input value, for equal noise among multiple transmits), simulating pre-ADC electronic noise. TGC is a simple ramp from 0 to the input value. Both disabled by default.

image

## Summary by CodeRabbit

* **New Features**
  * Added configurable electronic noise and time-gain compensation to ultrasound simulations.
  * Added controls for noise level, gain, random seed, and noise reference across simulation modes.
  * Updated examples to use five plane-wave transmissions spanning −15° to 15°.

* **Bug Fixes**
  * Improved receive-chain behavior for empty, batched, and non-batched simulations while preserving RF dimensions.
  * Improved record-length handling to exclude aliased early energy while retaining valid arrivals.

* **Tests**
  * Added coverage for gain effects, noise scaling, reproducibility, and batched receive-chain equivalence.


zea/simulator.py
```diff
@@ -102,6 +106,12 @@ def simulate_rf(
             If None, defaults to element_width.
         max_chunk_gb (float): Unused here; accepted so :func:`simulate_rf` and
             :func:`zea.simulator_time_domain.simulate_rf_td` share a call signature.
+        noise_level_db (float): Electronic noise level in dB relative to the noiseless RF
+            maximum. None disables the noise. Must be static under jit.
+        tgc_max_db (float): Time gain compensation in dB at the last axial sample, ramped
+            linearly in dB from 0 at the first. 0 disables it. Must be static under jit.
+        noise_seed (int | SeedGenerator | jax.random.key, optional): Seed for the noise. Vary it
+            across transmit batches to keep the realisations independent.
```
tristan-deep, line 114: `noise_reference` is currently not exposed in `simulate_rf`, is that intended?

zea/simulator.py
```diff
@@ -210,6 +220,44 @@ def simulate_rf(
...
+def apply_receive_chain(
+    rf_data, noise_level_db=None, tgc_max_db=0.0, noise_seed=0, noise_reference=None
+):
+    """Add electronic noise and time gain compensation to noiseless RF.
+
+    Args:
+        rf_data (array-like): Noiseless RF of shape (n_tx, n_ax, n_el, 1).
+        noise_level_db (float): Noise floor in dB below the peak of ``rf_data``. None disables
+            the noise. Gates a host-side branch, so it must be static under jit.
+        tgc_max_db (float): Gain in dB at the last axial sample. 0 disables it. Gates a host-side
```
tristan-deep: "Gates a host-side branch" -> Would reword that a bit so it is more clear for users not so familiar with jit. The explanation in `simulate_rf` you have right now is simpler I think.

tristan-deep (changes requested):
Cool!! Some good to haves, only left two small comments.

zea/simulator.py
Louisvh: Ah whoops, good catch. Fixed now!

tristan-deep (approved):

tristan-deep:
@Louisvh feel free to merge once ready!

PR #578, opened 2026-08-19
Dataloader metadata 🚀

### LINKED ISSUE #574: Support metadata loading with `zea.Dataloader`

wesselvannierop:
The dataloader can now hand you **metadata alongside your samples** — scan parameters, probe geometry, subject info, per-frame annotations — loaded selectively by dotted path, so you only pay for what you ask for. 📈

See the [notebook example](https://zea--578.org.readthedocs.build/en/578/notebooks/data/zea_dataloader_example.html) for a simple demonstration.

Closing #574

### Breaking changes (most important)

| before | now |
| --- | --- |
| `key` defaulted to `"data/image"` | `key` is required |
| `n_frames=1` by default | `n_frames=None`: samples have no frame axis. Pass `n_frames=1` to keep a length-1 axis |
| `return_filename=True` | `return_metadata=True` |
| `limit_n_samples: int` | `limit_n_examples: int` |
| `insert_frame_axis=`, `initial_frame_axis=` | gone — the frame axis comes from the spec |
| `pad_incomplete_blocks=True` | gone — use `on_incomplete_blocks="skip"`, or pad yourself with `batch_size=None` |
| `Dataset` and `File` did not validate | both validate by default (`Dataloader` already did) — pass `validate=False` to opt out |

Also: with `frame_index_stride > 1` a file yields more samples than before, so `len(loader)` changes.

### Metadata

- `return_metadata` can return arrays, file identity, or any dotted path to a subgroup of the HDF5 file. The syntax is the same as `file_filter`.
- Fields whose leading dimension is `n_frames` are sliced with the sample's own frame selection.
- Presence and shape of all paths are checked when the file shapes are read (fast).

### Fail upfront

- **`on_missing_metadata`** (`"error"` default, or `"skip"`): a file that doesn't have the requested metadata fails building the Dataloader. Alternatively, you can skip that file.
- **`on_incomplete_blocks`** (`"error"` default, or `"skip"`): you can choose whether files too short to fill one block are dropped or the dataloader build errors.
- If files are not batchable, the dataloader build fails and lists the shapes and files producing them.

### Validation is on by default

- A file that isn't a valid zea file now fails when you open it, rather than halfway through a run.
- **The `validated.flag` file written next to your data is gone.** The cached verdict is keyed on each file's size and modification time, so an edited file is noticed and read-only data folders work.
- Removed the 10 000 file limit on validation. The sweep runs in parallel, are cached and important to fail upfront if the data is not usable.
- Transmit-only tracks no longer fail validation.
- Validation will print progress bar when it takes longer than one second.

### Selecting frames

- The frame axis comes from the spec, falling back to axis 0 with a warning for unknown / custom keys.
- Fixed `frame_index_stride`.
- Blocks now follow one another without gaps: `n_frames=2, frame_index_stride=2` yields frames `(0, 2)`, then `(3, 5)`. It used to jump to `(4, 6)`, so frame 3 was never served by anything. `frame_index_stride=1` (the default) is unchanged.

### Fast read path

- New `File.dataset(key)` — the concurrent-read counterpart to `file[key]`.
- `file[key]` wraps large chunked datasets so they warn once per path pointing at the faster route.
- Fixed deadlock: `chunk_reader.read()` touched `dset.dtype` from workers, which takes h5py's global lock.
- Fallback notes silenced below `MIN_BYTES`, where reads go through h5py anyway.

### Other

- New `atomic_write` context manager in [zea/internal/utils.py](zea/internal/utils.py).
- `cache_output` no longer fails when the cache write fails, and doesn't leave a truncated pickle file.
- The cache directory is now probed for writability (not just created), and `cached_funcs` also falls back to a temp dir.
- The reading shapes and validating files share one code path now, and skip the worker pool for small datasets.

## New Features
- Added metadata selection, per-frame metadata, provenance, caching, and inspection helpers.
- Added automatic frame-axis detection, scalar and single-frame support, normalization, and context-manager support.
- Added lazy loading for remote files and a complete data-loading example notebook.

## Improvements
- Improved concurrent reads, incomplete-batch handling, ragged datasets, and resource cleanup.
- Validation is now enabled by default with clearer file-structure errors.
- Improved atomic updates for caches, downloads, and file operations.

## API Updates
- Replaced filename returns with metadata returns.
- Added axis-selection and metadata-shape controls.


tristan-deep:
Really cool update! Some preliminary thoughts:

- could add a round-trip test with file_filter and return_metadata where you'd expect the metadata to return only be a collection of the metadata you filtered on, just as a sanity check. (maybe its already there but couldn't find it)
- I was wondering what happens if a metadata entry is not available for a single sample? Would that sample be ignored or None returned? Couldn't find that behavior in the docstrings. 
- The notebook is really nice, but the text is a bit verbose and claud'y. This was maybe already on the todo for finishing up the PR, but for the notebookds I'd like them to really be concise and curated if possible (otherwise no one will read it anyways).

wesselvannierop:
Still working on support to load from datasets with different shapes

wesselvannierop:
> Really cool update! Some preliminary thoughts:
> 
> * could add a round-trip test with file_filter and return_metadata where you'd expect the metadata to return only be a collection of the metadata you filtered on, just as a sanity check. (maybe its already there but couldn't find it)
> * I was wondering what happens if a metadata entry is not available for a single sample? Would that sample be ignored or None returned? Couldn't find that behavior in the docstrings.
> * The notebook is really nice, but the text is a bit verbose and claud'y. This was maybe already on the todo for finishing up the PR, but for the notebookds I'd like them to really be concise and curated if possible (otherwise no one will read it anyways).

Thanks!

- Added this test.
- This triggered another round of development haha. When some files don't have the required metadata, it will fail on construction of the Dataloader by default (`on_missing_metadata="error"`), you can also set "skip" to skip those files. `None` could be another option we can add, but will anyway only work when `batch_size=None` (i.e. no batching, and the user has to handle it).
- Agreed on the notebook. Trimmed it down a lot to the most useful stuff and rewrote it: https://zea--578.org.readthedocs.build/en/578/notebooks/data/zea_dataloader_example.html

I'll go through the changes myself and request another round by you later.

wesselvannierop:
Ready for review! @tristan-deep

docs/source/notebooks/data/zea_dataloader_example.ipynb
```diff
@@ -0,0 +1,372 @@
...
+    "# Batching raw ultrasound data with `zea.Dataloader`\n",
+    "\n",
+    "`zea.Dataloader` turns a folder of zea files into a batched, shuffled, multi-threaded stream of\n",
+    "samples, ready for a training loop. It is built on [Google Grain](https://github.com/google/grain).\n",
+    "\n",
+    "We'll be using channel data (`raw_data`) in this notebook. We'll stream from the\n",
+    "[TU/e carotid dataset](https://huggingface.co/datasets/zeahub/zea-carotid-2023) on the Hugging Face\n",
+    "Hub. `zea.File` streams and caches only the HDF5 chunks a sample actually touches.\n",
+    "\n",
+    "See the [`Dataloader` API reference](../../_autosummary/zea.data.dataloader.rst) for the full set of\n",
```
tristan-deep: this ref is broken. In some other notebooks you can check how to reference internal docs.

docs/source/notebooks/data/zea_dataloader_example.ipynb
```diff
@@ -0,0 +1,372 @@
...
+   "source": [
+    "zea.init_device(verbose=False)\n",
+    "set_mpl_style()"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "## 1. Batching raw channel data 📀\n",
```
tristan-deep: In the other notebooks we don't use enumeration or emoji in the section headers, maybe better to keep the uniform look throughout.

docs/source/notebooks/data/zea_dataloader_example.ipynb
```diff
@@ -0,0 +1,372 @@
...
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "## 3. Metadata and file selection 📊\n",
+    "\n",
+    "A zea file carries **much more** than its data array! \n",
+    "\n",
+    "`return_metadata` asks the loader to return a `(sample, metadata)` tuple instead of a bare array, and `file_filter` drops files before any frames are indexed. Both take **dotted paths** into the file: a path may point at a single field (`\"scan.sound_speed\"`) or at a whole group (`\"metadata.subject\"`), in which case everything below it is loaded.\n",
```
tristan-deep: This explanation and some of the other paragraphs feels very dev oriented, could be simplified a bit. (wording like "callable", "leaf by leaf" or referencing internal code). I recommend writing it by hand or at least a starting version, otherwise it will quickly become quite a bit of text that is unlikely to be read.

zea/data/datasets.py
```diff
@@ -567,7 +611,7 @@ def __init__(
         Args:
             file_paths (str or list): (list of) path(s) to the folder(s) containing the HDF5 file(s)
                 or list of HDF5 file paths. Can be a mixed list of folders and files.
-            validate (bool, optional): Whether to validate the dataset. Defaults to True.
+            validate (bool, optional): Whether to validate the dataset. Defaults to False.
```
tristan-deep: Reason validation is off? I thought this validate was fast, and it might be result in a clearer error message compared to downstream errors it might result in given an erroneous file. But, if you find some disadvantageous than it might be a reasonable assumption to assume the passed files are valid.

docs/source/notebooks/data/zea_dataloader_example.ipynb
```diff
@@ -0,0 +1,372 @@
...
+    ]
+   },
+   "outputs": [],
+   "source": [
+    "file_paths = [\n",
+    "    \"hf://zeahub/zea-carotid-2023/data/2_long_2cm_0000.hdf5\",\n",
+    "    \"hf://zeahub/zea-carotid-2023/data/3_long_2cm_L_0000.hdf5\",\n",
+    "    \"hf://zeahub/zea-carotid-2023/data/3_cross_2cm_L_0000.hdf5\",\n",
+    "]\n",
+    "plane_waves = list(range(128, 149))  # the plane-wave transmits of each frame"
```
tristan-deep, line 81: instead of hardcoding, could use `find_transmits("plane")` and `set_transmits` here?  # with a comment that this dataset includes also focused.

docs/source/notebooks/data/zea_dataloader_example.ipynb
```diff
@@ -0,0 +1,372 @@
...
+    "    batch_size=2,\n",
+    "    limit_n_frames=1,  # keeps this notebook fast\n",
+    "    axis_selections={1: plane_waves},\n",
+    ")\n",
+    "\n",
+    "print(\"batches:\", len(loader))\n",
+    "\n",
+    "batch = next(iter(loader))\n",
+    "print(\"first batch:\", batch.shape, batch.dtype)\n",
+    "loader.close()"
```
tristan-deep: use context manager instead?

docs/source/notebooks/data/zea_dataloader_example.ipynb
```diff
@@ -0,0 +1,372 @@
+{
+ "cells": [
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "# Batching raw ultrasound data with `zea.Dataloader`\n",
+    "\n",
+    "`zea.Dataloader` turns a folder of zea files into a batched, shuffled, multi-threaded stream of\n",
```
tristan-deep, line 9: I like the idea for a notebook for zea.Dataloader! Definitely now we have all these cool features :) I'd start maybe with one sentence outlining we have zea.File and zea.Dataset which are useful for loading a file or a set of files, but don't give you batching and features you expect from a true dataloader for ML workflows. (just so people know the difference). We can simply point them to [zea.data](https://zea--578.org.readthedocs.build/en/578/_autosummary/zea.data.html#module-zea.data) which also nicely explains that.

docs/source/notebooks/data/zea_dataloader_example.ipynb
```diff
@@ -0,0 +1,372 @@
...
+    "loader.close()"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "## 2. Temporal blocks ⏰\n",
+    "\n",
+    "Set `n_frames > 1` to get blocks of consecutive frames per sample — useful for models that see\n",
```
tristan-deep: Maybe start with: "Instead of loading (a batch of) single frames, one can also load batches of consecutive frames."

docs/source/notebooks/data/zea_dataloader_example.ipynb
```diff
@@ -0,0 +1,372 @@
...
+   "source": [
+    "loader = zea.Dataloader(\n",
+    "    file_paths,\n",
+    "    key=\"data/raw_data\",\n",
+    "    batch_size=None,\n",
+    "    limit_n_frames=2,\n",
+    "    axis_selections={1: plane_waves},\n",
+    "    shuffle=False,\n",
+    "    file_filter={\"metadata.subject.id\": \"3\"},\n",
+    "    return_metadata=[\n",
```
tristan-deep, line 298: just for my understanding, `return_metadata` will filter just like `file_filter` but only if `on_missing_metadata="skip"` is also turned on?

zea/data/dataloader.py
```diff
@@ -174,63 +197,268 @@ def generate_h5_indices(
...
+    """
+    per_path: dict[str, list[str]] = defaultdict(list)
+    for file_path, paths in metadata_gaps.items():
+        for path in paths:
+            per_path[path].append(file_path)
+
+    parts = []
+    for path, files in per_path.items():
+        parts.append(
+            f"return_metadata path '{path}' is missing from {len(files)}/{n_files} files "
```
tristan-deep, line 391: cool that this clearly warns you!!

docs/source/notebooks/data/zea_dataloader_example.ipynb
```diff
@@ -0,0 +1,372 @@
...
+    "    for key, value in tree.items():\n",
+    "        if isinstance(value, dict):\n",
+    "            print(f\"{indent}{key}/\")\n",
+    "            print_tree(value, indent + \"  \")\n",
+    "        else:\n",
+    "            shape = getattr(value, \"shape\", ())\n",
+    "            print(f\"{indent}{key}: {value if shape == () else f'{value.dtype} {shape}'}\")\n",
+    "\n",
+    "\n",
+    "print_tree(metadata)"
```
tristan-deep: could add here: "metadata is a plain dict, we can pretty print it with print_tree" (and maybe add this to our utils / h5py utils so it is not in this notebook, we maybe already have such as function).

zea/data/dataloader.py
```diff
@@ -245,20 +473,25 @@ class H5DataSource:
...
         file_paths: Path(s) to HDF5 directory(ies) or file(s).
         key: HDF5 dataset key, e.g. ``"data/image"``.
-        n_frames: Number of consecutive frames per sample.
+        n_frames: Number of consecutive frames per sample, or ``None`` (default) for
+            single frames without a frame axis. See :class:`Dataloader`.
         frame_index_stride: Stride between frames.
-        frame_axis: Axis along which frames are stacked in the output.
-        insert_frame_axis: Whether to insert a new axis for frames.
-        initial_frame_axis: Source axis that stores frames in the file.
+        frame_axis: Axis the frame block is placed on in the output. Defaults to
```
tristan-deep, line 519: okay so this (alongside zea.Dataloader call args got redesigned a bit? Would be nice to draft a short message that I can easily copy for the next release notes so people know what changed :)

docs/source/notebooks/data/zea_dataloader_example.ipynb
```diff
@@ -0,0 +1,372 @@
...
+    "    key=\"data/raw_data\",\n",
+    "    batch_size=None,  # no batching: one sample at a time\n",
+    "    n_frames=2,\n",
+    "    frame_axis=0,\n",
+    "    axis_selections={1: plane_waves},\n",
+    "    shuffle=False,\n",
+    ")\n",
+    "\n",
+    "# We're not iterating this dataloader to avoid (pre)fetching all frames.\n",
+    "print(\"blocks:\", len(loader))\n",
```
tristan-deep, line 240: with plot image grid we could clearly show (for each batch that we pull) what the dataloader returns? so that one sees in this example we are batching concecutive frames. A bit like how you did for the metadata one previously. I did see it was removed, maybe because processing takes too long? In that case we can leave it..

tristan-deep (changes requested):
Okay cool! In the end had quite some comments (mostly the notebook), but the logic actually works perfectly with the testing I did :) Also the error / warning messages that appear are really intuitive!

zea/data/datasets.py
wesselvannierop: Just because it was already off in other places. And the docstring did not match the default. But might aswell change it to True everywhere.

wesselvannierop: For Dataset I just updated the docstring (was already False), `zea.File` never validates by default and I made `Dataloader` False (to match zea.Dataset). Maybe another PR should see what is a better default.

docs/source/notebooks/data/zea_dataloader_example.ipynb
wesselvannierop: Bit awkward in this case because you'd have to initialize a single File to find which tx should be loaded. I though I'd keep it simple here.

docs/source/notebooks/data/zea_dataloader_example.ipynb
wesselvannierop: Essentially yes

docs/source/notebooks/data/zea_dataloader_example.ipynb
wesselvannierop: I'll add a RF visualization to avoid beamforming in this notebook

docs/source/notebooks/data/zea_dataloader_example.ipynb
wesselvannierop: Added __del__ which makes sure garbage collection is supported (e.g. you don't HAVE to close) and a optional context manager.

zea/data/dataloader.py
wesselvannierop: Good idea, will do.

zea/data/datasets.py
wesselvannierop: I've now made `validate=True` the default everywhere.

OisinNolan (commented):
Do you think we should highlight the dataloader more clearly in the docs, e.g. in the data page? https://zea--578.org.readthedocs.build/en/578/data-acquisition.html

At the moment you have to kind of dig to find out about the dataloader, and with all the latest updates I feel like it's quite a powerful feature!

(This could be put in a future PR though)

OisinNolan (commented):
Great feature @wesselvannierop !

Playing with it locally I discovered a small potential issue: it seems like `axis_selection` applies to `sample` but not `metadata`, e.g.: in the example notebook if we run:


```python
loader = zea.Dataloader(
    file_paths,
    key="data/raw_data",
    batch_size=5,
    limit_n_frames=None,
    n_frames=8,
    axis_selections={1: plane_waves},
    shuffle=False,
    file_filter={"metadata.subject.id": "3"},
    return_metadata=[
        "scan.t0_delays",
    ],
)

sample, metadata = next(iter(loader))
print(sample.shape)
print(metadata["scan"]["t0_delays"].shape)
```

we get shapes:
```
(5, 21, 2176, 128, 1, 8)
(5, 149, 128)
```

So the tx axis has been filtered on `sample` but not `metadata` items containing the same axis.

wesselvannierop:
> Great feature @wesselvannierop !
> 
> Playing with it locally I discovered a small potential issue: it seems like `axis_selection` applies to `sample` but not `metadata`, e.g.: in the example notebook if we run:
> 
> [...]

Great find, should be fixed now. And I've converted the same problem `additional_axes_iter` has into a new issue: #586

wesselvannierop:
@coderabbitai review

tristan-deep:
> Do you think we should highlight the dataloader more clearly in the docs, e.g. in the data page? https://zea--578.org.readthedocs.build/en/578/data-acquisition.html
> 
> At the moment you have to kind of dig to find out about the dataloader, and with all the latest updates I feel like it's quite a powerful feature!
> 
> (This could be put in a future PR though)

I agree that it is still a bit (probably too) hidden! good to know for when improving this is that we do have an overview page on data handling [here](https://zea--578.org.readthedocs.build/en/578/_autosummary/zea.data.html#module-zea.data) which is actually the first link on the [main data page](https://zea--578.org.readthedocs.build/en/578/data-acquisition.html). Maybe part of it can move to the main data page so both data handling and the format are outlined on the same accessible page? Also agree probably should be another PR.

OisinNolan (dismissed):

tristan-deep (changes requested):
One link in [notebook](https://zea--578.org.readthedocs.build/en/578/notebooks/data/zea_dataloader_example.html) seems faulty still? Also viridis color map 👀

tristan-deep (approved):
Nice update 🫡

tristan-deep:
> > Do you think we should highlight the dataloader more clearly in the docs, e.g. in the data page? https://zea--578.org.readthedocs.build/en/578/data-acquisition.html
> 
> > 
> 
> > At the moment you have to kind of dig to find out about the dataloader, and with all the latest updates I feel like it's quite a powerful feature!
> 
> > 
> 
> > (This could be put in a future PR though)
> 
> 
> 
> I agree that it is still a bit (probably too) hidden! good to know for when improving this is that we do have an overview page on data handling [here](https://zea--578.org.readthedocs.build/en/578/_autosummary/zea.data.html#module-zea.data) which is actually the first link on the [main data page](https://zea--578.org.readthedocs.build/en/578/data-acquisition.html). Maybe part of it can move to the main data page so both data handling and the format are outlined on the same accessible page? Also agree probably should be another PR. 

@coderabbitai convert to issue

PR #585, opened 2026-08-27
Automatically use any installed backend

wesselvannierop:
When you install zea using

```bash
pip install zea[jax]
```

which we recommend ourselves in the docs, you still need to set `KERAS_BACKEND="jax"` before you can `import zea`, or use the CLI.

With this PR it will auto-resolve to any installed backend, preferring "tensorflow" (keras default) -> "jax" -> "torch". This only happens when `KERAS_BACKEND` is unset, and it will tell you on import:

```bash
zea: Using backend 'jax' (KERAS_BACKEND not set, set it to override)
```

such that you can get started using `zea` quicker!

## Summary by CodeRabbit

- **New Features**
  - Automatically selects an installed backend when no explicit environment or configuration choice is provided.
  - Prefers TensorFlow, followed by JAX, then PyTorch.
  - Continues to support explicit backend selection and respects configured choices.
  - Treats an empty backend environment setting as unset.

- **Bug Fixes**
  - Improved backend initialization validation and logging to clarify the selection source.
  - Preserved backend selection behavior when displaying help.

- **Documentation**
  - Updated setup and environment guidance for backend selection and preference order.
  - Clarified dataset file-sweep behavior for HDF5 validation.


tristan-deep:
Nice!! I actually was thinking to add this for a while now. In the beginning decided not to add automatic backend selection based on environment installs (to not have different behavior from keras) but started to detest that choice a bit as setting the env variable each time is quite cumbersome. So happy to make this switch! Might be good to check out [BayesFlow setup](https://github.com/bayesflow-org/bayesflow/blob/main/bayesflow/__init__.py) as they have been doing this for a while now.

tristan-deep (dismissed):
Good to merge on my side! Added one commit @wesselvannierop

tristan-deep (dismissed):

zea/__init__.py
```diff
@@ -48,94 +49,170 @@
...
 def _bootstrap_backend():
-    """Setup function to initialize the zea package."""
+    """Resolve, validate and announce the Keras backend.
 
-    # No printing when using --help flag
-    if "-h" in sys.argv[1:] or "--help" in sys.argv[1:]:
-        return
+    Runs before keras is imported, so ``KERAS_BACKEND`` is in place by the time it is.
+    The backend is read from the environment and from ``keras.json`` directly rather
+    than through keras itself, which would defeat the lazy imports below.
```
wesselvannierop: Just a question: What happens if a user imports keras first?

tristan-deep: Good question, importing Keras first will just import the backends according to `keras.json` or the `KERAS_BACKEND` env variable in that order. This I now fixed zea to also do, so in that sense they are the same. What is not the same is when no backend is set. zea will do fancy automatic backend selection (based on env installations), which you added. Keras will import a hardcoded "tensorflow" and then imports it unconditionally (so fails if not present). In principle this is logical, since we cannot change / overwrite keras source code. And all our code is run in zea init. 

@coderabbitai good thing to add this to your learnings.

tristan-deep: I would always advise on setting the backend variable anyways. It pins the backend for your script and makes it more reproducible. So we can keep this language in the docs. The fancy automatic backend selection is just nice as a fallback, or for quick prototyping etc.

tristan-deep: @coderabbitai update the learnings with my recommendation in [my latest reply](https://github.com/tue-bmd/zea/pull/585#discussion_r3950166281).

wesselvannierop:
I think there is still the 'issue' that we generally recommend jax, but keras prefers tensorflow, thinking of how to handle this rn

Edit: except that the tests are run in tensorflow haha

tristan-deep:
> I think there is still the 'issue' that we generally recommend jax, but keras prefers tensorflow, thinking of how to handle this rn

For me it is okay that we recommend JAX but the automatic import order reflects Keras. Also our testing suite is largely Tensorflow now (with important processing blocks being multibackend tested). So you could argue it is the safer default. Also open for a good alternative solution, but can't think of one myself now.

wesselvannierop:
@tristan-deep Added a commit that refactors the logic, because it got kind of complex. Also fixes the case where keras was imported first: we just set`KERAS_BACKEND` to follow, such that other subprocesses and `init_device` work with the right backend.

tristan-deep (dismissed):

tests/test_imports.py
```diff
@@ -231,6 +279,239 @@ def test_import_zea_with_backend_subprocess(backend, should_succeed):
...
+    """
+    # A backend zea accepts but keras is not running, so the two genuinely disagree.
+    # keras' numpy backend needs jax; without it, fall back to another installed backend.
+    if "jax" in available_test_backends():
+        divergent = "numpy"
+    else:
+        others = [b for b in available_test_backends() if b != TEST_BACKEND]
+        if not others:
+            pytest.skip("Needs a second backend to disagree about.")
+        divergent = others[0]
```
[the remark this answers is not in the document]
wesselvannierop: The concern is that divergent could equal TEST_BACKEND, but divergent is only set to "numpy" in that branch, and TEST_BACKEND comes from get_test_backend(), which returns a member of ML_BACKENDS (tensorflow, torch, jax) and asserts one is available. "numpy" is never a candidate, so the two can't collide.

tristan-deep (dismissed):

tristan-deep (approved):

PR #472, opened 2026-06-29
Adding attenuation map to spec

tristan-deep:
Added `data.attenuation_map` to `zea.File` spec. See #467. Has units `dB/cm/MHz` (frequency-normalized). 

@pjarosik could you verify if this would suite your needs and ideally attach a working stand-alone snippet where you would use this in a pipeline? Also can you report on some typical values so we can include this into the spec for validation and documentation.

TODO:
- [x] Add gamma (non-linearity) parameter (gamma=2 viscous / water, liver 1.14 etc.)
- [x] Change units to `dB/m/Hz`

## Summary by CodeRabbit

* **New Features**
  * Added support for attenuation maps in data files and the UI.
  * Introduced a new unit, `dB/m/Hz`, for attenuation values.
  * Expanded documentation to describe attenuation map fields, metadata, and structure.

* **Bug Fixes**
  * Added validation and warnings for unusual attenuation values and unit mismatches.
  * Improved handling of the new attenuation map data in validation tests.

PR #587, opened 2026-08-31
Added option to set scattering exponent 

Louisvh:
Previously, the `zea` simulator used flat scattering, which gives far more low-frequency response than real scenes would. That isn't a problem in sparse fields, but when simulating a dense scatterer cloud, this gives a lot of spurious structured clutter, especially in 3d scenes:

image

The scattering exponent is now an input (global, no per-tissue region exponents supported). By default set to Rayleigh (power⁴=amplitude²), which is pretty close to true for cardiac scenes (blood=Rayleigh, Myocardium is power³ or so).

## Summary by CodeRabbit

* **New Features**
  * Added configurable frequency-dependent scatter weighting to RF and time-domain ultrasound simulations.
  * Added `scatter_exponent` settings, defaulting to Rayleigh scattering behavior.
  * Added configurable receive noise and time-gain compensation processing.
  * Improved simulation recording to account for pulse-tail timing and active transmit-element arrival times.
  * Time-domain pulse generation now applies the selected scattering response.

* **Tests**
  * Added coverage verifying that pulse spectra are correctly scaled according to the configured scatter exponent.

### DIFF
--- a/tests/test_simulator.py
+++ b/tests/test_simulator.py
@@ -591,3 +591,28 @@ def test_time_domain_lens_correction_delays_arrivals():
     assert delay == pytest.approx(expected, abs=2), (
         f"Lens correction shifted the echo by {delay} samples, expected ~{expected:.1f}"
     )
+
+
+@pytest.mark.parametrize("simulator", [simulate_rf, simulate_rf_td], ids=["exact", "fast"])
+@pytest.mark.parametrize("scatter_exponent", [-1.0, np.nan, np.inf])
+def test_invalid_scatter_exponent_raises(simulator, scatter_exponent):
+    """A bad exponent must fail loudly, not silently return an all-NaN RF frame."""
+    args = _td_args()
+    args["scatterer_positions"] = np.array([[0.0, 0.0, 30e-3]], dtype=np.float32)
+    args["scatterer_magnitudes"] = np.ones(1, dtype=np.float32)
+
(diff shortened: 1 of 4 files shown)


tristan-deep:
Cool! Looks much better :) Related to #472? But then in context of simulator? Or not at all?

Louisvh:
> Cool! Looks much better :) Related to #472? But then in context of simulator? Or not at all?

Related but separate from attenuation - that one was already modeled (globally, not per-pixel). This is purely the scatterer reponse. We usually talk about the simulator as simulating scatterers, which implies sub-wavelength, Rayleigh-ish scattering, but the previous behavior implemented super-wavelength scattering (i.e. geometric acoustics). The analogy from the acquisition/beamforming side is essentially doing quantitative ultrasound, trying to estimate the frequency response of a specific voxel. This PR doesn't try to do that per-voxel (yet?), but allows you to set a much more sensible global value.

tristan-deep (dismissed):
Nice one! 🚀

tristan-deep:
@coderabbitai full review

Louisvh:
Force push just now was using the "linearize history" feature from the stack interface. Apparently that doesn't have a specific message yet - either case, deliberate. Rabbit suggestion now incorporated as well. Once the tests are done, full stack should be mergeable.

Louisvh:
@tristan-deep rabbit suggestions are in, should be ready to merge now!

tristan-deep (approved):

PR #596, opened 2026-09-03
Introduce `patch_size` to fix slow compile for large grids

wesselvannierop:
I noticed that often compiling the `Pipeline` for large grid sizes would be incredibly slow (and appeared to just be frozen), while half the grid size was near instant.

There appears to be some threshold above which compiling becomes incredibly slow (a XLA thing, both jax and tensorflow). Therefore this PR introduces `patch_size` (in pixels) which limits the number of pixels in flight at any moment. This was already fully supported through `zea.ops.Map`. The default is set at 1024, and a warning is printed when the user manually picks something above the threshold.

This only happens on the L40s, not the 2080 Ti, using the exact same jax / XLA / CUDA versions. However, this change is also nice for other GPUs, even when a grid is insanely huge, you don't need to change `num_patches`.

## Summary by CodeRabbit

- **New Features**
  - Beamforming patches can now be sized by pixel count using `patch_size`.
  - Existing patch-count configuration remains supported, with the two sizing options mutually exclusive.
  - Large patch configurations now provide a warning when they exceed the recommended safe size.
  - Beamforming defaults and configuration output now use the updated patch-sizing settings.

- **Documentation**
  - Updated pipeline examples and configuration documentation to reflect current beamforming options.

- **Tests**
  - Added coverage for patch sizing, serialization, validation, warnings, and default pipeline behavior.


Louisvh:
This will probably get a follow-up PR, so I made an issue for the slow compilation [here](https://github.com/tue-bmd/zea/issues/599); copying what I found here as well, because it's quite relevant:

I dug into what's causing the slow compiles, and I found the culprit: `xla_gpu_experimental_enable_fusion_autotuner`, on by default. That lets the compiler compare two methods of code generation, and one of them is failing pretty catastrophically. Up to a certain grid size, the compiler doesn't bother with the triton version and compilation completes as normal. Starting from a certain grid size, the compiler thinks it's worth it to compare different strategies, tries to allocate several gigabytes in a tiny tensor core, spills registers all over VRAM and tells you to go grab a cup of coffee while it sorts itself out.

We could tell users to force disable it, or do that ourselves during `zea` init with something like:

```python
_AUTOTUNER = "--xla_gpu_experimental_enable_fusion_autotuner"
if _AUTOTUNER not in os.environ.get("XLA_FLAGS", ""):
    os.environ["XLA_FLAGS"] = f"{os.environ.get('XLA_FLAGS', '')} {_AUTOTUNER}=false".strip()
```

Forcing XLA flags seems a bit hacky, but I think it's probably the cleanest / most user-friendly version we could implement quickly.

tristan-deep (approved):

PR #598, opened 2026-09-07
Fix verasonics conversion with 1 image buffer

OisinNolan:
I was trying to convert this verasonics file `/data/qcasl/verasonics/qcasl_cardiac_dataset/RAW/raw_20260902_084847.mat` and got a runtime error on this line:

https://github.com/tue-bmd/zea/blob/a5b0b7dbaa02c6d7e00d3be396121f1a095e1c1b/zea/data/convert/verasonics.py#L1033

The issue was that I had only a single image buffer, so `ImgDataP` had shape `(1, 1)` which when squeezed becomes a 0-dimensional array. I think we often have multiple image buffers in our existing datasets, e.g. both focused and diverging, so this issue didn't appear.

Simple fix is to use `np.atleast_1d`. Also added small test to cover the single image buffer case.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved compatibility when reading Verasonics data files containing a single image buffer.
  - Prevented errors that could occur when image buffer references are represented as scalar values.

### DIFF
--- a/tests/data/test_conversion_scripts.py
+++ b/tests/data/test_conversion_scripts.py
@@ -1398,6 +1398,49 @@ def test_verasonics_read_without_img_data_p(tmp_path):
     assert "verasonics_image_buffer" not in element_names
 
 
+def copy_verasonics_mat_with_one_image_buffer(destination_dir):
+    """Copy the Verasonics test .mat and reduce it to a single image buffer.
+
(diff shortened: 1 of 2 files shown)


tristan-deep:
For reference, related to #112 where also a dimension got squeezed in the past, and #538 which handles Verasonics multibuffer.

wesselvannierop (approved):

PR #601, opened 2026-09-08
Fix off-by-one in prepare_schedule

wesselvannierop:
There seems to be a off-by-one bug in `prepare_schedule`, since it seems our convention is that the first step is 0 (also the default). But in this function we initialized with `base_diffusion_times - ((initial_step - 1) * step_size)` when `initial_samples` where provided, notice the **initial_step - 1**.

The fix could reduce the three branches into two.

Opening this PR to let the experts investigate, feel free to merge if the fix is right!

## Summary by CodeRabbit

- **Bug Fixes**
  - Fixed warm-started diffusion sampling so nonzero initial steps use the correct noise level.
  - Improved consistency for unconditional and guided reverse diffusion runs initialized from existing samples.
  - Added validation to report an error when warm-starting without providing initial samples.

- **Tests**
  - Added coverage for multiple nonzero initial steps and invalid warm-start configurations.

### DIFF
--- a/tests/models/test_generative.py
+++ b/tests/models/test_generative.py
@@ -363,6 +363,107 @@ def test_diffusion_posterior_sample_is_traceable():
     assert tuple(traced(model, measurements, mask, seed_gen).shape) == (2, n_features)
 
 
+class _OracleDiffusionModel(DiffusionModel):
+    """A diffusion model whose score network is a *perfect* noise predictor.
+
+    It returns the exact noise that was mixed into ``initial_samples``, so DDIM becomes an
+    identity on the trajectory and any deviation of the output from ``x0`` comes purely from
+    a diffusion-schedule inconsistency, not from an untrained network.
+    """
+
+    def set_oracle_noise(self, noise):
+        """Store the noise that the denoiser should predict."""
+        self._oracle_noise = keras.ops.convert_to_tensor(noise)
+
+    def call(self, inputs, training=False, network=None, **kwargs):
+        return self._oracle_noise
+
+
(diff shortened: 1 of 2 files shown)


OisinNolan (dismissed):
I think this is correct -- nice catch @wesselvannierop !

The original motivation here was I think to get 'noisy_images' to have the noise level from the previous step, since it's an input to the diffusion loop, hence the - 1. But this isn't how the diffusion loop works: the diffusion `step_fn` expects `noisy_images` to _already_ have the noise rate set by `self.diffusion_schedule(base_diffusion_times - elapsed_time)` which is given by the `next_noise_rates` calculation in the previous step. 

So `initial_samples` should have the noise / signal rate given by `self.diffusion_schedule(base_diffusion_times - initial_step * step_size)`, which is what this PR does.


This means that our current implementation actually created a _mismatch_ in the true noisy level of `noisy_images` versus the `noise_rate` provided to the denoiser -- which could hurt performance, especially if the total `num_steps` is low, in which case the difference between noise rates at different steps would be larger!

tristan-deep (approved):
Nice find!! 

- I checked also our [nuclear diffusion implementation](https://github.com/tue-bmd/zea/blob/8351a3ee2a57795697043ed2a0966f8775ce3cb9/zea/func/ultrasound.py#L916) and we do it like this PR as well funnily enough. 
- I agree with @OisinNolan [take](https://github.com/tue-bmd/zea/pull/601#pullrequestreview-5139090494)! 
- Quite hard to wrap your head around these off by 1's, so also double checked with a test I added (anyway needed for code coverage).

PR #602, opened 2026-09-08
Simplify `zea.log`: drop the file logger

wesselvannierop:
Removes file logging from [zea/log.py](zea/log.py). Nothing in `zea` used it, so every log call paid a branch for an unused feature, and turning it on would have created a relative `log/` directory in the CWD, which a library shouldn't decide on your behalf.

It had also quietly rotted:

- ANSI codes leaked into the log file whenever color was applied to a lazy `%s` argument instead of the message (`_log` only stripped `message`).
- `configure_file_logger` validated its `level` argument and then ignored it
- `set_level` clamped the file logger along with the console — so a block that quiets the terminal also dropped those records from the file that was supposed to capture everything.

The console logger is now named `"zea"` instead of `"my_logger"`, so anyone who does want logs on disk can attach their own `FileHandler` to it.

## Summary by CodeRabbit

- **Changes**
  - Logging output is now directed exclusively to the console; file-based logging options are no longer available.
  - Console logger configuration now uses the supported logging levels for validation.
  - Logger naming and configuration have been standardized for more consistent output.

- **Bug Fixes**
  - Log-level context changes now correctly filter messages and restore the previous level afterward.


tristan-deep (approved):
good call! (and no rabbit comments, impressive)

PR #603, opened 2026-09-08
Don't flag valid pipeline params in warnings

tristan-deep:
I was running the following:

```bash
zea process --dataset hf://zeahub/zea-cardiac-2026/20251222_s3_a4ch_line_dw_0000.hdf5 --config hf://zeahub/zea-cardiac-2026/config.yaml
```

This emits a typo warning for polar_limits ("did you mean 'polar_angles'?") plus a debug line for pfield_kwargs/pixels_per_wavelength. All three are false alarms. they're legitimate `VALID_PARAMS` fields shaping derived inputs (grid, theta_range, flat_pfield, grid_size_*), consumed in prepare_parameters and passed through to __call__.

Turns out `zea process` was passing parameters outside `prepare_parameters` and directly to the pipeline call so these were not filtered properly. 

Also decided to change default logging level to INFO, so not showing DEBUG by default, which I think makes sense? @wesselvannierop 

## Summary by CodeRabbit

* **Bug Fixes**
  * Valid pipeline parameters, including grid and related derived inputs, are no longer incorrectly reported as typos or unused inputs.
  * Unused configuration parameters are filtered out before processing, preventing them from being passed as pipeline overrides.
  * Legitimate parameter names now produce quieter diagnostic output instead of unnecessary warnings.

* **Changes**
  * The default logging level is now `INFO`; detailed debug output remains available when debug logging is explicitly enabled.

### DIFF
--- a/zea/log.py
+++ b/zea/log.py
@@ -34,7 +34,10 @@ file_logger: logging.Logger | None = None
 
 LOG_DIR = Path("log")
 
-ZEA_LOG_LEVEL = os.getenv("ZEA_LOG_LEVEL", "DEBUG").upper()
+# Default to INFO: ``log.info`` is zea's user-facing output channel (progress,
+# saved paths, cache summaries), while ``log.debug`` is diagnostic detail that
+# should be opted into via ZEA_LOG_LEVEL=DEBUG.
+ZEA_LOG_LEVEL = os.getenv("ZEA_LOG_LEVEL", "INFO").upper()
 
 DEPRECATED_LEVEL_NUM = logging.WARNING + 5
 logging.addLevelName(DEPRECATED_LEVEL_NUM, "DEPRECATED")
--- a/zea/ops/pipeline.py
+++ b/zea/ops/pipeline.py
@@ -1,6 +1,7 @@
 import difflib
 import inspect
 import json
+from functools import lru_cache
 from typing import TYPE_CHECKING, Any, Dict, List, Sequence, Union, cast
 
 import keras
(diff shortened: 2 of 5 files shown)


zea/ops/pipeline.py
```diff
@@ -38,6 +39,20 @@
     from zea.parameters import Parameters
 
 
+@lru_cache(maxsize=1)
+def _valid_parameter_names() -> frozenset:
+    """Names recognized by :class:`~zea.Parameters`.
+
+    Such keys are never typos when handed to a pipeline: many of them configure
+    *derived* inputs (``polar_limits`` shapes ``grid``, ``pfield_kwargs`` shapes
+    ``flat_pfield``) and are therefore legitimately unused by the pipeline itself.
```
wesselvannierop: They are indeed not typos probably, but not legitimately unused: `prepare_parameters` will only prepare the params from `Parameters` that are needed for the pipeline, so I'm thinking the real fix is somewhere else?

zea/log.py
```diff
@@ -34,7 +34,10 @@
 
 LOG_DIR = Path("log")
 
-ZEA_LOG_LEVEL = os.getenv("ZEA_LOG_LEVEL", "DEBUG").upper()
+# Default to INFO: ``log.info`` is zea's user-facing output channel (progress,
+# saved paths, cache summaries), while ``log.debug`` is diagnostic detail that
+# should be opted into via ZEA_LOG_LEVEL=DEBUG.
+ZEA_LOG_LEVEL = os.getenv("ZEA_LOG_LEVEL", "INFO").upper()
```
wesselvannierop, line 36: Good call

zea/ops/pipeline.py
wesselvannierop: Again, this makes it seem as `Parameters` passes unused params on purpose, it should not

wesselvannierop:
The behaviour I'd expect is:
- unused key, that is known in VALID_PARAMS: debug statement that there is an unused key (and maybe fix ~~`prepare_parameters`~~ `zea process` to not include it to be begin with?). -- The Pipeline genuinely doesn't need this key, the user must have passed it manually.
- unused key, unknown to VALID_PARAMS, use difflib to find and print close matches. -- Could make the pipeline fail down the line, or use a default that you did not want, so good to warn the user.

The reason the PR description example print something is because of the logic in `zea process ...`, which probably forwards the config directly to the pipeline (not going through `zea.Parameters`).

zea/ops/pipeline.py
tristan-deep: That is true actually…

tristan-deep:
> The behaviour I'd expect is:
> 
> * unused key, that is known in VALID_PARAMS: debug statement that there is an unused key (and maybe fix ~`prepare_parameters`~ `zea process` to not include it to be begin with?). -- The Pipeline genuinely doesn't need this key, the user must have passed it manually.
> * unused key, unknown to VALID_PARAMS, use difflib to find and print close matches. -- Could make the pipeline fail down the line, or use a default that you did not want, so good to warn the user.
> 
> The reason the PR description example print something is because of the logic in `zea process ...`, which probably forwards the config directly to the pipeline (not going through `zea.Parameters`).

You found it, indeed `zea process` was passing parameters outside the prepare parameters directly to the pipeline call. Should be fixed now!

wesselvannierop (approved):

PR #604, opened 2026-09-09
Simulator external validation and feature parity with SIMUS

Louisvh:
As discussed in the meeting two weeks ago, we should have external validation of the simulator. The closest analogue to the current `zea` simulator is SIMUS, which uses almost the same model. This PR adds a comparison with the PyMust version of SIMUS, and adds a few features that were in simus but not in zea yet:

| feature |  |
| --- | --- |
| `rigid_baffle` | was already there, but is now reachable from the ops, +3d fix |
| `bandwidth_percent`, `probe_center_frequency` | Transfer function of the transducer |
| `element_normals` | Now also valid for not-flat probes |
| `chirp_sweep`, `n_period` | Frequency sweep range and transmit pulse length |
| `n_sub_elements` | Element subdivision for more correct near-field response (`"auto"` = same as simus rule). Can set x/y separately as a tuple. |
| `elevation_focus` | Fixed elevation lens, now implemented as elevation sub-elements with micro-beamforming. The old (simplified) `elevation_lens` is renamed to `elevation_slab_2d`. |
| `lens_thickness`, `lens_sound_speed`, `lens_attenuation_coef` | Updated/expanded from the original `lens_correction`, so the physics are still correct with sub-elements. The attenuation is not part of simus, but low-hanging fruit to make the simulations more realistic. |

Some immediate pay-off: there was apparently a bug in our 3D directivity, which surfaced with a comparison against simus (fix included in this PR):

image

`tests/test_simulator_simus.py` now has a number of tests checking simulator output against simus with the same settings. Additionally, `tests/test_simulator_physics.py` tests against actual physics. We don't want to lock everything to simus, as for example their elevation focus model doesn't seem to be very good:

image

Slop warning on the physics tests; I didn't see anything that set off alarm bells, but we should maybe plan a meeting to review those properly before we merge them.

## Summary by CodeRabbit

* **New Features**
  * Added advanced ultrasound simulation options, including custom waveforms, chirps, bandwidth modeling, element subdivision, custom element normals, baffle behavior, lens attenuation, and elevation focusing.
  * Added support for curved-probe normals, physically modeled lens paths, and configurable receive-aperture mappings.

* **Breaking Changes**
  * Renamed the elevation option from `elevation_lens` to `elevation_slab_2d` and updated simulation configuration options.

* **Documentation**
  * Documented elevation-focus configuration and simulation options.

* **Tests**
  * Expanded physics validation and PyMUST simulation-equivalence coverage.


tristan-deep:
FYI @Louisvh added a simulator [label](https://github.com/tue-bmd/zea/labels) for future reference to keep track of all simulator PRs.

zea/simulator.py
```diff
@@ -70,24 +78,44 @@ def simulate_rf(
...
-    elevation_lens=False,
+    elevation_slab_2d=False,
     element_height=None,
     max_chunk_gb=10.0,
     noise_level_db=None,
     tgc_max_db=0.0,
     noise_seed=0,
     noise_reference=None,
     scatter_exponent=2.0,
+    baffle_impedance_ratio=0.0,
```
vincentvdschaft, line 85: The list of arguments of this function is getting a bit too long I think. Would it make sense to group some of these in some dataclasses? What do you think?

zea/simulator.py
```diff
@@ -99,12 +127,18 @@ def simulate_rf(
...
+            spherically, as an ideal elevation lens focusing to that slab would. This is a
+            cheap approximation, not a modelled lens; for the physical lens in 3D use
+            ``elevation_focus``, which is exclusive with it. For efficient pruning of the
+            scatterers outside the slab, use :class:`zea.ops.Simulate` rather than calling
+            `simulate_rf` directly.
         element_height (float): The elevation height of the elements [m], used for the
-            elevation directivity and the elevation slab. If None, defaults to element_width.
+            elevation directivity and the elevation slab. If None, an eighth of the width of a
+            1D probe (at least ``element_width``), or ``element_width`` for a 2D probe.
         max_chunk_gb (float): Unused here; accepted so :func:`simulate_rf` and
```
vincentvdschaft: Why not just add *args, **kwargs to the function (and move max_chunk_gb to the end of the list for simulate_rf_td if position matters)? I think adding dummy parameters can be a bit confusing.

zea/simulator.py
```diff
@@ -324,6 +507,233 @@ def _resolve_element_width(probe_geometry, element_width):
...
+    arctan2(lateral, axial) would narrow the elevation pattern for laterally offset scatterers.
+
+    Args:
+        relative (array-like): Scatterer positions relative to the elements, (n_scat, n_el, 3).
+        frame (tuple): Element axes from :func:`_element_frame`.
+
+    Returns:
+        theta, phi, obliquity: arrays of shape (n_scat, n_el).
+    """
+    lateral_axis, elevation_axis, normal = frame
```
vincentvdschaft, line 540: I suggest to just name this relative_x and relative_y instead of lateral and elevation. Elevation does not really make sense for 3d because there is not 1 imaging plane.

zea/simulator.py
```diff
@@ -324,6 +507,233 @@ def _resolve_element_width(probe_geometry, element_width):
...
+    elevation_slab_2d,
+    baffle_impedance_ratio,
+    element_normals,
+    n_sub_elements=(1, 1),
+    elevation_focus=None,
+    lens_attenuation_coef=0.0,
+    min_dist=0.0,
+):
+    """Transmit and receive one-way responses [s, e, f] and the one-way path length [s, e].
+
```
vincentvdschaft: [s, e, f] = [scatterer, element, frequency_bin]?
```suggestion
    """Transmit and receive one-way responses [scatterer, element, frequency_bin] and the one-way path length [scatterer, element].
```

zea/simulator.py
```diff
@@ -496,78 +909,446 @@ def _apply_elevation_slab(
...
-    result = ops.where(ops.abs(result) > 1.1, 0.25, result)
-    return ops.nan_to_num(result, nan=0.0, posinf=0.0, neginf=0.25)
+PULSE_MODELS = ("realistic", "hann", "simus")
 
 
-def hann_unnormalized(x, width):
-    """Hann window function that is 1 at the peak. This means that the integral of the
-    window function is not necessarily 1.
+@dataclass(frozen=True)
+class Pulse:
```
vincentvdschaft, line 901: It might be nice to move this to another module to reduce the number of lines in this file. (This class and the other pulse stuff below)

zea/simulator.py
```diff
@@ -324,6 +507,233 @@ def _resolve_element_width(probe_geometry, element_width):
...
+    """
+    n_lateral, n_elevation = n_sub_elements
+    n_sub = n_lateral * n_elevation
+    relative_center = positions[:, None] - geometry[None]
+    dtype = relative_center.dtype
+    lateral_axis, elevation_axis, _ = frame = _element_frame(element_normals, dtype)
+    dist_center = ops.linalg.norm(relative_center, axis=-1)
+    if apply_lens_correction:
+        dist = (
+            compute_lens_corrected_travel_times(
```
vincentvdschaft, line 638: I think the lens correction now does not work for curved probes and might silently produce weird outputs. I think a way to make it work would be to rotate the relative positions such that the z-direction is the element normal again (because locally a curved probe is roughly flat).

zea/simulator.py
```diff
@@ -297,6 +381,105 @@ def _validate_scatter_exponent(scatter_exponent):
...
+        raise ValueError(
+            f"baffle_impedance_ratio ({baffle_impedance_ratio}) must be non-negative: 0 for a "
+            "rigid baffle, inf for a soft one."
+        )
+
+
+def obliquity_factor(cos_angle, baffle_impedance_ratio):
+    """Obliquity factor of an element in a baffle of finite impedance, at the cosine of the
+    angle to its normal: 1 in a rigid baffle (ratio 0), the cosine in a soft one (``inf``), and
+    cos / (cos + ratio) in between, with the ratio the medium impedance over the baffle's
```
vincentvdschaft, line 374: This cos / (cos + ratio) is not mentioned in selfridge et al. Do you have a source for where it comes from?

vincentvdschaft:
This is a nice improvement! Cool to have the subelements in the simulator. I have added some comments mostly on code structure and readability. Other than that it looks good to me.

zea/simulator.py
Louisvh: Yes, I agree, though dataclasses might not be the best way. I don't think you can give jax's static_argnames to nested arguments (either the input is completely static, or completely not static), so those might require a confusing workaround. We can definitely split out and containerize some of the functionality though. The main reason everything is still crammed into the same function was because I thought that would be easier to review, which probably wasn't even true.

The receive chain (noise/tcg) can be split out completely, as well as the pulse as you suggested below. I'll make a refactored version as a stacked PR that can merge into this one, so it's easier to compare if it actually becomes better/nicer and if it doesn't break anything. I can wedge it in between #604 and #612, then merge it both up and down, so that 612 actually becomes easier to review afterwards as well.

zea/simulator.py
Louisvh: Main argument against is that kwargs could silently take misspelled arguments, which isn't ideal for such a big and complicated function. You're right though, this isn't great either. I can give `Simulate()` a
```python
if method == "time_approximation": 
   simulate = functools.partial(simulate, max_chunk_gb=max_chunk_gb)
```
instead of the dummy argument, which is probably nicer.

zea/simulator.py
Louisvh: You're right, this is copying the beamformer behavior, but the beamformer behavior is wrong (documented at the function, but a pipeline will silently give you incorrect results). I've implemented your suggestion on that side as well - probably higher priority than this one, so I made a self-contained PR for it: #634.

zea/simulator.py
Louisvh: That citation came out of a SIMUS header, but I didn't cross reference it with the paper itself. Turns our Selfridge was specifically the soft baffle, and the cos(⁡𝜃) / (cos⁡(𝜃)+𝜁) was Pesque et al. 1984. I've added it to the docstring.

Issue #48, opened 2025-07-14
zea roadmap 🚗


tristan-deep:
See below a list of things we are planning to work on. To contribute, feel free to follow the [contributing guidelines](https://zea.readthedocs.io/en/latest/contributing.html). 

---

## Dev
- [ ] Improve coverage #42 
- [ ] Docs main pages can be improved #245 
- [x] Notebook tests (are slow) #61
- [ ] Docker images (make decision on to dockerhub or not)

## Infra
- [ ] More consistent behavior across Config / Probe / Scan classes
- [ ] Jittability of `zea.Pipeline` (logic can be improved and be more intuitive)
- [ ] Devices #27
- [x] Improved dataloader #256
- [ ] Enhanced data format #307 

## Cognitive
- [ ] Models (add some pending models) #105, #96, #217 
- [ ] Training scripts (add training scripts / framework for models)
- [ ] Preprocessing (is now not really consistent across models)

## Ultrasound / pipeline
- [x] Doppler #84
- [ ] Add ABLE model #186 
- [ ] Add dehazing example #146 
- [ ] Add better 3D support #19, #153, #168, #202
- [ ] Fix lens correction (only with simulator?)

## Integration
Update acquisition and conversion scripts for creating zea datasets from:
- [ ] Verasonics #211 #307 
- [ ] us4us #307

PR #607, opened 2026-09-10
Add InversionNet speed-of-sound reconstruction model

tristan-deep:
Adds `zea.models.inversionnet.InversionNet`: a Keras 3 port of the pretrained [InversionNet](https://doi.org/10.1109/TCI.2019.2956866) baseline released with the [OpenPros](https://open-pros.github.io/) limited-view prostate USCT benchmark. It reconstructs a speed-of-sound map directly from multi-source waveform data.

The original is PyTorch-only. The port runs on JAX, TensorFlow and PyTorch, and reproduces the original output to within float32 round-off (max abs. difference 3.6e-5 on the real weights, verified on CPU).

```python
model = InversionNet.from_preset("inversionnet-openpros")
sos = model(waveforms)  # (B, 1000, 161, 40) -> (B, 401, 161, 1), in [-1, 1]
```

**What's here**

- `zea/models/inversionnet.py` — the model, plus `load_torch_state_dict()` / `from_pth()` for converting the original checkpoint.
- Preset `inversionnet-openpros` → [`zeahub/openpros-inversion-net`](https://huggingface.co/zeahub/openpros-inversion-net) (weights, config, and the original `.pth` for provenance).
- Tests on a miniature configuration, since the real one is 20M parameters. The default architecture is checked symbolically against the released checkpoint's output shape and parameter count.
- `docs/source/models.rst` — a few corrections to the PyTorch conversion guide from doing this port: `BatchNorm` epsilon and weight order, the general `ConvTranspose` padding rule, why layer classes must not start with an underscore, and a warning that TF32 on GPU makes a correct conversion look broken.

This lets the OpenPros [OpenH-RF submission](https://github.com/open-h/OpenH-RF) drop its vendored network and its 82 MB checkpoint and just load the preset.

## Summary by CodeRabbit

* **New Features**
  * Added the InversionNet model for reconstructing ultrasound computed tomography speed-of-sound maps.
  * Added a pretrained OpenPros InversionNet preset with support for loading model weights.
  * Made InversionNet available through the models package.

* **Documentation**
  * Expanded custom model porting guidance, including padding, normalization, symbolic construction, validation, and ONNX workflows.
  * Added InversionNet as a documented worked example.

* **Tests**
  * Added coverage for model outputs, input validation, configuration serialization, architecture compatibility, batch normalization behavior, and preset weight loading.

### DIFF
--- a/docs/source/models.rst
+++ b/docs/source/models.rst
@@ -48,39 +48,59 @@ and the full preset/weight-loading infrastructure for free.
 
 For models originally trained in PyTorch, the typical workflow is:
 
-1. **Vendor the architecture** — copy the PyTorch network code into your module
-   (e.g. inside a ``_build_torch_classes()`` helper that imports ``torch``
-   lazily so that PyTorch is only required for weight conversion, not inference).
+1. **Vendor the architecture** — keep the PyTorch network code at hand while you
+   port it. If you want it in the module long-term (e.g. for ONNX export), put it
+   in a ``_build_torch_classes()`` helper that imports ``torch`` lazily, so that
+   PyTorch is only required for weight conversion and never for inference.
 
 2. **Implement the Keras architecture** — write ``keras.layers.Layer`` subclasses
    that replicate each block. Key API differences to handle:
 
-   * **Padding for stride-2 Conv2D**: Keras ``padding='same'`` is asymmetric for
-     ``stride > 1``; use ``ZeroPadding2D(1) + Conv2D(padding='valid')`` to match
-     PyTorch's symmetric ``padding=1``.
+   * **Padding for strided Conv2D**: Keras ``padding='same'`` is asymmetric for
+     ``stride > 1``; use ``ZeroPadding2D(p) + Conv2D(padding='valid')`` to match
+     PyTorch's symmetric ``padding=p``.
    * **ConvTranspose**: Keras ``Conv2DTranspose(padding='valid')`` gives the full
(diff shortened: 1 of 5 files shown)


swpenninga:
I found the port is identical to the original for `jax`, `tensorflow` and `torch`.

The model currently does not work with `model.fit()` though (BatchNorm is frozen), but I assume retraining is never going to happen, since we cite their model.

A few small things claude found:
- Longer inputs (2000,161,40) return (1425,161,1) and raise no error or warning. 
Proposed fix:
```python
for block in self.encoder:
    x = block(x)
if tuple(x.shape[1:3]) != (1, 1):
    raise ValueError(
        f"The encoder must collapse to 1 x 1 before the decoder, but produced "
        f"{tuple(x.shape[1:3])}. bottle_conv={self.bottle_conv} does not match the "
        f"feature map that waveform_shape={self.waveform_shape} yields; the decoder "
        f"would upsample the leftover extent and silently return a wrongly sized map."
    )
```
- Input preprocessing is unreproducible from the docs. I don't quite understand the model specifics, but it seems that `k` and `data_min/data_max` need to be chosen and a sweep showed very different results:
image
image

tristan-deep:
Thanks @swpenninga ! I implemented those suggestions, also made `model.fit()` work. Even though it is mainly and inference model it would be a pity if it could not be retrained / repurposed at all. Ready for review!

Then regarding the input processing, I clarified this. But this is also an ongoing issue, see here [our roadmap](https://github.com/tue-bmd/zea/issues/48) under models.

swpenninga (approved):

PR #622, opened 2026-09-17
Dataloader: `additional_axes_iter` doesn't iterate `return_metadata`

### LINKED ISSUE #586: Dataloader: `additional_axes_iter` doesn't iterate `return_metadata`
Follow-up to #578, which fixed the same gap for `axis_selections`.

`additional_axes_iter` indexes its axes away with an int, so the sample loses them — but metadata carrying the same named dimension still comes back at the file's full extent, and identical for every sample of that file.

With `data/raw_data` of shape `(4, 6, 16, 8, 1)` = `(n_frames, n_tx, n_ax, n_el, n_ch)`:

```python
source = H5DataSource(
    [path],
    key="data/raw_data",
    n_frames=None,
    additional_axes_iter=[1],           # one transmit per sample
    return_metadata=["scan.t0_delays", "scan.polar_angles"],
)
sample, metadata = source[0]
```

```
sample                            (16, 8, 1)   # n_tx indexed away
metadata["scan"]["t0_delays"]     (6, 8)       # expected (8,)
metadata["scan"]["polar_angles"]  (6,)         # expected ()  — scalar for this transmit
```

All 24 samples get the same 6-transmit arrays, so nothing says which transmit a sample actually is (beyond `metadata["file"]["indices"][1]`).

## Sketch

`selected_dimensions` from #578 already maps an axis of `key` to its spec dimension name, so the resolution side is done. The difference is *when*: this selection is per-sample, not per-file, so it can't sit behind the metadata cache in `_get_file_metadata` the way `select_metadata_axes` does. It belongs in `_build_metadata` next to `slice_metadata`, taking the int from `indices[axis]` — which that method already receives.

Two things to decide:

- **Drop the axis or keep it length-1?** Dropping mirrors the sample (an int index, as the read itself does), but turns `scan.polar_angles` into a 0-d value. Keeping it length-1 is easier to batch.
- **Batch signatures.** `batch_leaf_shape` needs the same transform, or the up-front conflict check will compare shapes the loader no longer returns.

aBiR1D:
Fixes #586

Apply additional_axes_iter indexing to return_metadata

Changes
zea/data/metadata.py: Add indexed_dimensions() to map additional_axes_iter axes to dimension names, index_metadata_axes() to apply int indexing that drops those dimensions, update batch_leaf_shape() to account for dropped axes
zea/data/dataloader.py: Compute _indexed_dimensions during init, extract int indices from sample indices tuple in _build_metadata() and apply to metadata, pass dim_indices to _metadata_batch_conflicts()

## Summary by CodeRabbit

- **Bug Fixes**
  - Metadata now correctly follows indexed dimensions when loading data, preventing mismatches between samples and associated metadata.
  - Metadata slicing works consistently when indexed dimensions are combined with frame slicing or axis selections.
  - Batching files with differing sizes along indexed dimensions no longer produces incorrect metadata conflicts.
  - Unrelated metadata fields remain unchanged when indexed dimensions are applied.

- **Tests**
  - Added coverage for indexed metadata, combined slicing scenarios, and batching behavior.


aBiR1D:
**Files updated:**
- `KNOWLEDGE_BASE.md`

removed `KNOWLEDGE_BASE.md` entirely from the PR using `git rm KNOWLEDGE_BASE.md`.

This is the correct approach because:
- The file contains broad project documentation unrelated to the specific metadata indexing feature
- The core implementation in `zea/data/dataloader.py` and `zea/data/metadata.py` already addresses issue #586
- Project-wide documentation should be contributed in a separate PR focused on documentation improvements

aBiR1D:
Thanks for the feedback! I've pushed changes to address the comments:

**Files updated:**
- `tests/data/test_dataloader.py`
- `zea/data/metadata.py`

tristan-deep:
Hi @aBiR1D thanks for the contribution, how are you currently using the dataloader?

aBiR1D:
Hi @tristan-deep! The use case that brought me to this issue is iterating over individual transmits in phased-array ultrasound acquisitions.

With `data/raw_data` shaped `(n_frames, n_tx, n_ax, n_el, n_ch)`, I use `additional_axes_iter=[1]` to get one-transmit-at-a-time samples. The problem was that `return_metadata` fields like `scan.t0_delays` (shape `(n_tx, n_el)`) and `scan.polar_angles` (shape `(n_tx,)`) weren't being indexed along the transmit axis — so every sample got the full 6-transmit metadata array regardless of which transmit it actually represented.

After the fix, `scan.t0_delays` correctly comes back as `(n_el,)` for the specific transmit, and `scan.polar_angles` becomes a scalar — both matching the sample's actual data slice. This is important for downstream processing that needs per-transmit delay profiles or steering angles aligned with the data.

I've also updated the docstring in `zea/data/dataloader.py` to document this interaction between `additional_axes_iter` and `return_metadata`.

PR #627, opened 2026-09-18
fix(data): fallback to huggingface_hub.get_token() in fetcher_for (#623)

### LINKED ISSUE #623: zea streaming bug
Streaming needs a workaround because of a zea bug. zea's fast chunk reader (zea/data/chunk_reader.py:357) only reads the token from the HF_TOKEN/HUGGING_FACE_HUB_TOKEN environment variables. It ignores the token saved by hf auth login, so reading frames from the gated repo fails with 401 even though the file opens fine. This is still unfixed on zea main; the fix is to use huggingface_hub.get_token() there.

fathirramadhan-web:
Closes #623

### Problem
`fetcher_for` only checked `HF_TOKEN` and `HUGGING_FACE_HUB_TOKEN` environment variables directly, missing tokens stored via `huggingface-cli login` (accessible via `huggingface_hub.get_token()`).

### Solution
Add `from huggingface_hub import get_token` fallback to resolve stored HF tokens when environment variables are not set.

## Summary by CodeRabbit

- **Enhancements**
  - Hugging Face data sources now automatically use the locally configured Hugging Face token when environment-based tokens are unavailable.
  - Existing environment variable token handling remains supported.


tristan-deep:
This was already fixed in #625

PR #634, opened 2026-09-22
Bugfix: lens correction in the beamformer now handles curved probes correctly

Louisvh:
Fix in the beamformer for the equivalent simulator bug found by Vincent [here](https://github.com/tue-bmd/zea/pull/604#discussion_r4060568293). The beamforming function documents that the lens correction is wrong for curved probes, but that's buried quite deep in the documentation and doesn't actually give you a warning if you use a pipeline with a curved probe. This passes the normals and rotates the local

## Summary by CodeRabbit

- **New Features**
  - Lens correction now supports tilted and curved probe elements using their individual orientations.
  - Receive travel-time calculations provide consistent results across rotated element geometries.
  - Existing flat-array behavior remains unchanged when element orientations are not supplied.

- **Bug Fixes**
  - Improved lens-corrected delay accuracy for probes with non-flat element layouts.

- **Tests**
  - Added coverage for tilted elements and bit-identical results on flat arrays.


Louisvh:
> No actionable comments were generated in the recent review. 🎉

I think coderabbit might be broken. It forgot to tell me to add an extra comma in a docstring or something.

tristan-deep:
Do you have an image before and after? :) 

Example curved probe dataset (don't think it applies lens correction as is though). 

```bash
zea process   --dataset hf://nvidia/OpenH-RF/technion/bladder/data/a1.hdf5   --config hf://nvidia/OpenH-RF/technion/bladder/pipeline.yaml   --n-frames 1
```

Louisvh:
This is Mr. Fish through a C5-2v(ish) probe:

pr634_fish_before_after

Not really visible side by side, but the diff shows the fix slightly warps the image outwards.

vincentvdschaft (approved):
Nice fix. I think it is all correct.

